PR #4, opened 2023-12-17
Add documentation

dfed:
After a few weeks‘ of work, we have a working repository, and now: a readme!


ahmdmhasn (approved):

README.md
```diff
@@ -1,2 +1,422 @@
...
+    @Instantiated(fulfilledByType: "ChildView")
+    private let childViewBuilder: Instantiator<some View>
+}
+```
+
+### @Forwarded
+
+Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Forwarded` macro](Sources/SafeDI/PropertyDecoration/Forwarded.swift) are forwarded into the SafeDI dependency tree by a [`ForwardingInstantiator`](Sources/SafeDI/DelayedInstantiation/ForwardingInstantiator.swift) instance’s `instantiate(…)` method. A `@Forwarded`-decorated property is available to be [`@Received`](#received) by objects instantiated further down the dependency tree.
+
+A single `@Instantiable` type may have at most one `@Forwarded`-decorated property.
```
kierajmumick, line 216: I think we probably should provide a matrix illustrating the situations when one should use instantiated, forwarded, and received. It would make it clear at a glance what to use in what scenarios

kierajmumick (commented):
All looks pretty good! I have some comments and suggestions!

README.md
```diff
@@ -1,2 +1,422 @@
...
+
+## Using SafeDI
+
+SafeDI utilizes Swift Macros and Swift Package Manager plugins to read your code and generate a dependency tree that is validated at compile time. Dependencies can either be instantiated by SafeDI or forwarded into the SafeDI dependency tree.
+
+Opting a type into the SafeDI dependency tree is straightforward: add the `@Instantiable` macro to your type declaration, and decorate your type‘s dependencies with macros that signal the lifecycle of each property. Decorate a property with `@Instantiated` if you want to initialize it when the enclosing type is initialized; `@Forwarded` if you want to propagate a runtime-determine value down the dependency tree; or `@Received` if you want to receive the property from an `@Instantiated` or `@Forwarded` property further up the dependency tree.
+
+If a type is declared in third-party code, you can declare an extension the type in your code and decorate it with the `@ExternalInstantiable` macro to opt it into SafeDI.
+
+Let‘s walk through each of these macros in detail.
```
ahmdmhasn, line 44: The inclusion of graphs enhances clarity regarding the various types. Without prior familiarity with the potency of dependency injection in the tree, comprehending this concept initially might prove challenging.

ahmdmhasn: Also, I would suggest to mention the different Macros ahead and then jump into an example of each.

ahmdmhasn: The intro is fine in this part

README.md
```diff
@@ -1,2 +1,422 @@
...
+        securePersistentStorage["user", ofType: User.self]
+    }
+
+    private func persistUserData() {
+        securePersistentStorage["user"] = user
+    }
+}
+```
+
+### @ExternalInstantiable
```
ahmdmhasn, line 121: I think you should keep the same order as in the intro, so `@ExternalInstantiable` should come after `@Received`

README.md
```diff
@@ -1,2 +1,422 @@
...
+
+/// A default implementation of `SecurePersistentStorage` that can fulfill
+/// `@Instantiated` properties of type `Valet` or `SecurePersistentStorage`.
+@ExternalInstantiable(fulfillingAdditionalTypes: [SecurePersistentStorage.self])
+extension Valet: SecurePersistentStorage {
+
+    /// A public initializer defines how SafeDI can instantiate a Valet object.
+    public static func instantiate() -> Valet {
+        Valet.valet(
+            with: Identifier(nonEmpty: "SafeDIExample")!,
```
ahmdmhasn, line 151: Is it possible for Macros to deduce the identifier from the type name? Since most protocols typically have a sole default concrete implementation, adopting this approach would significantly enhance the readability of the API. 

Also, to facilitate understanding for beginners, I would recommend extending the earlier comment by simplifying the example. Specifically, suggesting a single method without parameters, such as public static func instantiate() -> Type, would make its clarity and purpose evident to the reader.

README.md
```diff
@@ -1,2 +1,422 @@
...
+                try? removeObject(forKey: key)
+            }
+        }
+    }
+}
+```
+
+### @Instantiated
+
+Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Instantiated` macro](Sources/SafeDI/PropertyDecoration/Instantiated.swift) are instantiated when its enclosing type is instantiated. `@Instantiated`-decorated properties are available to be [`@Received`](#received) by objects instantiated further down the dependency tree.
```
ahmdmhasn, line 178:
```suggestion
Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Instantiated`](Sources/SafeDI/PropertyDecoration/Instantiated.swift) are instantiated when its enclosing type is instantiated. `@Instantiated`-decorated properties are available to be [`@Received`](#received) by objects instantiated further down the dependency tree.
```
It has been defined earlier as a Macro, no need for repeating.

README.md
```diff
@@ -1,2 +1,422 @@
...
+    // for this code to compile is for there to be an
+    // `@Instantiable public struct ChildView: View` in the codebase.
+    @Instantiated(fulfilledByType: "ChildView")
+    private let childViewBuilder: Instantiator<some View>
+}
+```
+
+### @Forwarded
+
+Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Forwarded` macro](Sources/SafeDI/PropertyDecoration/Forwarded.swift) are forwarded into the SafeDI dependency tree by a [`ForwardingInstantiator`](Sources/SafeDI/DelayedInstantiation/ForwardingInstantiator.swift) instance’s `instantiate(…)` method. A `@Forwarded`-decorated property is available to be [`@Received`](#received) by objects instantiated further down the dependency tree.
```
ahmdmhasn, line 214:
```suggestion
Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Forwarded`](Sources/SafeDI/PropertyDecoration/Forwarded.swift) are forwarded into the SafeDI dependency tree by a [`ForwardingInstantiator`](Sources/SafeDI/DelayedInstantiation/ForwardingInstantiator.swift) instance’s `instantiate(…)` method. A `@Forwarded`-decorated property is available to be [`@Received`](#received) by objects instantiated further down the dependency tree.
```

README.md
```diff
@@ -1,2 +1,422 @@
...
+
+A single `@Instantiable` type may have at most one `@Forwarded`-decorated property.
+
+### @Received
+
+Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Received` macro](Sources/SafeDI/PropertyDecoration/Received.swift) are injected into the enclosing type‘s initializer. Received properties must be [`@Instantiated`](#instantiated) or [`@Forwarded`](#forwarded) by an object higher up in the dependency tree.
```
ahmdmhasn, line 220:
```suggestion
Property declarations within [`@Instantiable`](#instantiable) types decorated with the [`@Received`](Sources/SafeDI/PropertyDecoration/Received.swift) are injected into the enclosing type‘s initializer. Received properties must be [`@Instantiated`](#instantiated) or [`@Forwarded`](#forwarded) by an object higher up in the dependency tree.
```

README.md
```diff
@@ -1,2 +1,422 @@
...
+1. The type‘s SafeDI-injected properties are all `@Instantiated`
+2. The type is not instantiated by another `@Instantiable` type
+
+### Comparing SafeDI and Manual Injection: Key Differences
+
+SafeDI is designed to be simple to adopt and minimize architectural changes required to get the benefits of a compile-time safe DI system. Despite this design goal, there are a few key differences between projects that utilize SafeDI and projects that don‘t. As the benefits of this system are clearly outlined in the [Features](#features) section above, this section outlines the pattern changes required to utilize a DI system like SafeDI.
+
+#### Instantiating objects
+
+In a manual DI system, it is common to directly call your dependencies‘ `init(…)` methods. When utilizing SafeDI, you must rely on `@Instantiated`-decorated properties to instantiate your dependencies for you. Calling a dependency‘s `init(…)` method directly effectively exits the SafeDI-built dependency tree.
```
cyrilchandelier, line 308: > When utilizing SafeDI, you must rely on `@Instantiated`-decorated properties to instantiate your dependencies for you.
What happens if you don't? Say you start passing around a `@Instantiated` dependency your class received, will this cause any issue?

README.md
```diff
@@ -1,2 +1,422 @@
...
+### Creating the root of your dependency tree
+
+SafeDI automatically finds the root(s) of your dependency tree, and creates an extension on each root that contains a `public init()` method that instantiates the dependency tree.
+
+An `@Instantiable` type qualifies as the root of a dependency tree if and only if:
+
+1. The type‘s SafeDI-injected properties are all `@Instantiated`
+2. The type is not instantiated by another `@Instantiable` type
+
+### Comparing SafeDI and Manual Injection: Key Differences
```
ahmdmhasn, line 302: To highlight its strength, a useful method is to compare it with other Dependency Injection (DI) frameworks. Presenting the same key points in a table format, using X and ✔️, can effectively illustrate this comparison.

README.md
```diff
@@ -1,2 +1,422 @@
...
+@Instantiable(fulfillingAdditionalTypes: [UserService.self])
+public final class DefaultUserService: UserService {
+
+    // MARK: Initialization
+
+    /// Public, memberwise initializer that takes each injected property.
+    public init(authService: AuthService, securePersistentStorage: SecurePersistentStorage) {
+        self.authService = authService
+        self.securePersistentStorage = securePersistentStorage
+    }
```
cyrilchandelier, line 85: Any chance this initializer too can be generated?

README.md
```diff
@@ -1,2 +1,422 @@
...
+}
+```
+
+It is possible to write a `Instantiator` with a type-erased generic by utilizing `@Instantiated`‘s `fulfilledByType` parameter.
+
+#### ForwardingInstantiator
+
+The `ForwardingInstantiator` type is how SafeDI enables instantiating any `@Instantiable` type with a `@Forwarded` property. `ForwardingInstantiator` has two generics. The first generic must match the type of the `@Forwarded` property. The second generic matches the type of the to-be-instantiated instance.
+
+```swift
```
ahmdmhasn, line 256: Can we add another simple example in UIKit for an App level injection as well?

README.md
```diff
@@ -1,2 +1,422 @@
...
+
+✅ Easy testing: every type has a memberwise initializer
+
+✅ Clear error messages: never debug generated code
+
+## Using SafeDI
+
+SafeDI utilizes Swift Macros and Swift Package Manager plugins to read your code and generate a dependency tree that is validated at compile time. Dependencies can either be instantiated by SafeDI or forwarded into the SafeDI dependency tree.
+
+Opting a type into the SafeDI dependency tree is straightforward: add the `@Instantiable` macro to your type declaration, and decorate your type‘s dependencies with macros that signal the lifecycle of each property. Decorate a property with `@Instantiated` if you want to initialize it when the enclosing type is initialized; `@Forwarded` if you want to propagate a runtime-determine value down the dependency tree; or `@Received` if you want to receive the property from an `@Instantiated` or `@Forwarded` property further up the dependency tree.
```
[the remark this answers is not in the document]
dfed: fair. I reached for semicolons because these clauses are _long_ but... that's not a great reason 😅. I'll clean this up.

README.md
```diff
@@ -1,2 +1,422 @@
...
+import SafeDI
+
+/// A protocol that defines a UserService.
+/// It is not necessary to utilize protocols with SafeDI, but since
+/// protocol-driven development aids both testability and dependency
+/// inversion, our examples show how protocols can be used with SafeDI.
+public protocol UserService {
+    var user: User? { get }
+    func login(username: String, password: String) async throws -> User
+}
```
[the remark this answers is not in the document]
dfed: solid insight. could have one example without a protocol and another subsection+example with a protocol. starting with the simpler example makes sense.

README.md
```diff
@@ -1,2 +1,422 @@
...
+/// protocol-driven development aids both testability and dependency
+/// inversion, our examples show how protocols can be used with SafeDI.
+protocol SecurePersistentStorage {
+    subscript<CodableType: Codable>(_ key: String, ofType type: CodableType.Type) -> CodableType? { get }
+    subscript<CodableType: Codable>(_ key: String) -> CodableType? { get set }
+}
+
+/// A default implementation of `SecurePersistentStorage` that can fulfill
+/// `@Instantiated` properties of type `Valet` or `SecurePersistentStorage`.
+@ExternalInstantiable(fulfillingAdditionalTypes: [SecurePersistentStorage.self])
```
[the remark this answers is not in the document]
dfed: I hadn't thought about combining them! The fact that the implementations/capabilities are quite different doesn't really matter much from a "do we need a new macro" point of view. Certainly doable to combine. I'll give that a shot. Fewer macros would likely make the library simpler to understand.

README.md
dfed: Extremely smart. We're in need of better visual organization of this README for sure, and a table would be a solid step forward here.

README.md
```diff
@@ -1,2 +1,422 @@
...
+
+    /// A private property that knows how to instantiate a content view.
+    @Instantiated
+    private let contentViewInstantiator: Instantiator<ContentView>
+}
+```
+
+It is possible to write a `Instantiator` with a type-erased generic by utilizing `@Instantiated`‘s `fulfilledByType` parameter.
+
+#### ForwardingInstantiator
```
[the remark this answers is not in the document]
dfed: Yes! If you have a `@Forwarded` property you _must_ use a `ForwardingInstantiator` to instantiate the thing. Sounds like I need to make that more clear.

README.md
dfed: Per Kieraj above I'm likely to collapse `@ExternalInstantiable` into `@Instantiable` which will simplify the ordering a bit.

README.md
dfed: Identifiers are a Valet-specific thing... I should probably use the `UserDefaults` example from the example project rather than inventing a newer, more complicated example here 😅. Good thinking that this is a bit more complex than it needed to be.

README.md
dfed: I have a `Comparing SafeDI to other DI libraries` section, but it's pretty light-weight. I like the idea of a table here, but I'm not sure we need to go that hard just yet. This section is more of an adoption guide.

README.md
dfed: You can call `init` yourself, but then you're basically exiting the SafeDI dependency tree and there are no guarantees re when something is instantiated. Sounds like I should make that more clear here 🙂

README.md
dfed: In the simple case, it can! I've got [a TODO in the codebase to simplify this](https://github.com/dfed/SafeDI/blob/76b713a766232f203f72e1eeb303d01a81d69cfc/Sources/SafeDIMacros/Macros/InstantiableMacro.swift#L83).

When you have non-injected properties that must be set in `init`, though, you'll still need to write your own initializer. I like that today we can generate the structure of the initializer for you with a FixIt – if I started generating the initializer for you we'd lose the FixIt. I haven't figured out if the trade-off of simplifying the simple case while making the more complex (but still common!) case a bit more annoying is worth it yet.

Of course, now that I've typed that out... I could generate the FixIt only if you have a local property with an initializer... that'd be the best of both worlds. We’ll Get There™. Just gotta put in the time 🙂

README.md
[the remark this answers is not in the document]
kierajmumick: Could Forwarded be a property wrapper (instead of a macro) that automatically wraps the forwarded type in a `ForwardingInstantiator`?

README.md
```diff
@@ -1,2 +1,422 @@
...
+/// inversion, our examples show how protocols can be used with SafeDI.
+public protocol UserService {
+    var user: User? { get }
+    func login(username: String, password: String) async throws -> User
+}
+
+/// A default implementation of `UserService` that can fulfill `@Instantiated`
+/// properties of type `UserService` or `DefaultUserService`.
+@Instantiable(fulfillingAdditionalTypes: [UserService.self])
+public final class DefaultUserService: UserService {
```
kierajmumick, line 77: I know we talked about this probably about 100 times, and I know you disagree due to differences in coding style across the industry, but seeing this in front of me really does make me want to say this again 😂 

I really do think we should be able to infer that `XYZProtocolName: ProtocolName` fulfills the type `ProtocolName`, _unless_ `fulfillingAdditionalTypes` is provided as an argument.

Otherwise we're writing effectively boilerplate (We're writing the string `UserService` 3 times), which feels like something Macros should eliminate.

README.md
dfed: We can, but honestly I have significantly less-strong opinions about the right way to define a root using SafeDI in UIKit-world. 

Because `AppDelegate` and `SceneDelegate` must be separate objects, and neither object has a good way to reference the other (outside of the scene delegate referencing and casting `UIApplication.shared.delegate`)... there's no _great_ way to make an application root like there is in `SwiftUI`.

In my own spikes I've been resorting to making a
```
@Instantiable
public final class Root {

    public init(...) { ... }

    static let shared = Root() // uses initializer generated by SafeDI

    // Instantiated properties here
    ...
}
```

And then utilizing `Root.shared` in both the `AppDelegate` and `SceneDelegate`. This is a pretty kludgy (but workable) approach, and is similar to how folk tend to share objects between these two types anyways. Alternatively, you could have separate dependency trees for your `AppDelegate` and `SceneDelegate`... but that seems like a recipe for pain since scenes and the app do need to share context.

So... yeah. I didn't want to have an opinion on this in the README, since there is no great option. Does that make sense?

I agree that having a UIKit example will aid folk who are adopting this system in UIKit, but also... it's not hard to make the leap from "how do I share dependencies between these two delegates" to "I should create a self-vending singleton".

README.md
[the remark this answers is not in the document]
dfed: sorry, the `ForwardingInstantiator` is how you instantiate the _enclosing type_ that has a forwarded property, not the forwarded property itself. No way to really make this magical, especially because the `ForwardingInstantiator` needs the `@Forwarded` property as an argument in its `instantiate(...)` method.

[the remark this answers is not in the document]
kierajmumick: from what I remember, the macro is only used as a syntax flag for the plugin, so making this a property wrapper should work.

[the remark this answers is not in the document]
kierajmumick: if i'm understanding this right

README.md
dfed: We disagree on the point of a macro! The point isn't eliminating boilerplate as much as making it easier to do a standard thing.

At a high level: DI systems are already quite magical, so I really want to avoid "we made this assumption, didn't have a good way to tell you that we did so, and gave you no way to opt out without changing your type name" scenarios.

Imagine we had:
```
public final class ViewController: UIViewController { ... }

@Instantiable
public final class ProfileViewController: ViewController { ... }
```

I kinda doubt we would want to make `ProfileViewController` also fulfill `ViewController`. And if `ViewController` were `@Instantiable` the decision tree gets even more complex. Of course, we could come up with some set of rules to handle these cases and document them here, but... this kind of magic would make the learning curve steeper for this library.

The learning curve for adopting automated DI is already kinda reasonably high. I do not want to make it more complicated when the alternative is "repeat yourself in a way that makes it very clear what you're doing".

README.md
MrAdamBoyd: If someone forgets to add `@Instantiated` to a property, is this something they will notice easily?

README.md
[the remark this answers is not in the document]
dfed: oh yeah I've played with utilizing property wrappers instead of macros and it does work (I did this – and then deleted – for `@LazyInstantiated`).

Are you saying you would have something like
```
@Instantiated // macro
@Forwarded<ForwardArgument> // property wrapper
let builder: Instantiator<InstantiatedType>
```

Because we can't make `@Forwarded` itself a property wrapper that exposes a `ForwardingInstantiator` because the forwarding instantiator is the thing we need to _create the object with the forwarded property_.

All that said: what problem are we solving here?

README.md
dfed: short answer: yes they will notice quickly and easily.

If they fail add `@Instantiated`, `@Received`, or `@Forwarded` to a property on an `@Instantiatable` type, then SafeDI will fail the build (with a FixIt) until they write an `init` method that does not include this property as an argument. If they accept the FixIt, they'll need to instantiate the property themselves in `init(...)`, which should be a sign that they're not using DI for this property.

README.md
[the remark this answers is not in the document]
kierajmumick: okay I misunderstood—I thought you were saying this needed to be:
```
let builder: FordwardingInstantiator<InstantiatedType>
```

But considering that it is not, property wrappers don't make sense here. 

I was less trying to solve a problem, and more trying to understand how this works.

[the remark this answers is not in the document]
dfed: Got it. We need:
```
let builder: FordwardingInstantiator<ForwardedArgument, InstantiatedType>
```

The code example a few lines down hopefully makes this a bit more clear... but I'm very much hearing I need more English explanation of this requirement too.

README.md
dfed: 😅 can you tell which sections I wrote first? yeah I can simplify here.

README.md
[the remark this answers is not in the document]
dfed: Just took a pass at unifying in 0574004a5e3901707ad00790e5cac03644b8c7d8. Still need to update the docs here.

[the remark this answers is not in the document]
dfed: Docs have been updated!

README.md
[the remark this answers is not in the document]
dfed: Dramatically simplified the README examples

README.md
dfed: Added a "macro cheat sheet". The table in this section copied wording from elsewhere but I think the table representation is still helpful as an easy reference. LMK what you think.

README.md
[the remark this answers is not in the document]
dfed: I updated the `@Forwarded` description to make this a bit more clear.

README.md
dfed: See the new cheat sheet section, and the updated "Using SafeDI" section. I think I've addressed what you're getting at. Let me know!

README.md
dfed: I've now answered Cyril's q in this section!

README.md
```diff
@@ -1,2 +1,406 @@
...
+2. Have a `public init(…)` or `open init(…)` that receives every injectable property
+
+The `@Instantiable` macro guides engineers through satisfying these requirements with build-time FixIts.
+
+Here is a sample `UserService` implementation that is `@Instantiable`:
+
+```swift
+import SafeDI
+
+public final class UserService {
```
kierajmumick, line 72: I think this is missing an `@Instantiable` macro 😂

README.md
```diff
@@ -1,2 +1,406 @@
...
+
+/// A protocol that defines a UserService.
+public protocol UserService {
+    var user: User? { get }
+    func login(username: String, password: String) async throws -> User
+}
+
+/// A default implementation of `UserService` that can fulfill `@Instantiated`
+/// properties of type `UserService` or `DefaultUserService`.
+@Instantiable(fulfillingAdditionalTypes: [UserService.self])
```
kierajmumick, line 131: This is fufilling **additional** types because `DefaultUserService` will be used to fulfill dependencies of type `UserService` _and_ `DefaultUserService`. Is that right? Asking to make sure I understand clearly.

README.md
```diff
@@ -1,2 +1,406 @@
...
+import SecurePersistentStorage // A third-party library that provides secure, persistent storage.
+
+@Instantiable
+extension SecurePersistentStorage {
+    /// A public static function that defines how SafeDI can instantiate the type.
+    public static func instantiate() -> SecurePersistentStorage {
+        SecurePersistentStorage()
+    }
+}
+```
```
kierajmumick, line 155: Nice!! this looks great.

README.md
dfed: Oh my god. Simplified this example a little too much didn't I.

dfed:
```suggestion
@Instantiable
public final class UserService {
```

README.md
dfed: Correct!

kierajmumick (approved):
Looks great to me!

README.md
dfed: alright undid the cheat sheet sectioned put in a table at the top.

dfed:
Going to leave this PR open for a few more days to see if I can get another review pass with the new changes. Once this gets merged I'll release a 0.1.0 beta version of SafeDI!

README.md
```diff
@@ -1,2 +1,407 @@
...
+1. The type‘s SafeDI-injected properties are all `@Instantiated`
+2. The type is not instantiated by another `@Instantiable` type
+
+### Comparing SafeDI and Manual Injection: Key Differences
+
+SafeDI is designed to be simple to adopt and minimize architectural changes required to get the benefits of a compile-time safe DI system. Despite this design goal, there are a few key differences between projects that utilize SafeDI and projects that don‘t. As the benefits of this system are clearly outlined in the [Features](#features) section above, this section outlines the pattern changes required to utilize a DI system like SafeDI.
+
+#### Instantiating objects
+
+In a manual DI system, it is common to directly call your dependencies‘ `init(…)` functions. When utilizing SafeDI, you must rely on `@Instantiated`-decorated properties to instantiate your dependencies for you. Calling a dependency‘s `init(…)` function directly effectively exits the SafeDI-built dependency tree, which removes property lifecycle guarantees.
```
NickEntin, line 293: With the exception of the root, right?

README.md
```diff
@@ -1,2 +1,407 @@
...
+
+`@Instantiated`, `@Forwarded`, or `@Received` objects may be decorated with [`@ObservedObject`](https://developer.apple.com/documentation/swiftui/ObservedObject). Note that `@Instantiated` objects declared on a `View` will be re-initialized when the view is re-initialized. You can find a deep dive on SwiftUI view lifecycles [here](https://www.donnywals.com/understanding-how-and-when-swiftui-decides-to-redraw-views/).
+
+### Migrating to SafeDI
+
+It is strongly recommended that projects adopting SafeDI start their migration by identifying the root of their dependency tree and making it `@Instantiable`. Once your root object has adopted SafeDI, continue migrating dependencies to SafeDI in either a breadth-first or depth-first manner. As your adoption of SafeDI progression, you‘ll find that you are removing more code than you are adding: many of your dependencies are likely being passed through intermediary objects that do not utilize the dependency except to instantiate a dependency deeper in the tree. Once types further down the dependency tree have adopted SafeDI, you will be able to avoid receiving dependencies in intermediary types.
```
NickEntin, line 313:
```suggestion
It is strongly recommended that projects adopting SafeDI start their migration by identifying the root of their dependency tree and making it `@Instantiable`. Once your root object has adopted SafeDI, continue migrating dependencies to SafeDI in either a breadth-first or depth-first manner. As your adoption of SafeDI progresses, you‘ll find that you are removing more code than you are adding: many of your dependencies are likely being passed through intermediary objects that do not utilize the dependency except to instantiate a dependency deeper in the tree. Once types further down the dependency tree have adopted SafeDI, you will be able to avoid receiving dependencies in intermediary types.
```

README.md
```diff
@@ -1,2 +1,407 @@
...
+
+Type declarations decorated with [`@Instantiable`](Sources/SafeDI/PropertyDecoration/Instantiable.swift) are able to be instantiated by SafeDI. Types decorated with this macro can instantiate other `@Instantiable` dependencies, forward dependencies injected from outside of SafeDI, or receive dependencies instantiated or forwarded by objects further up the dependency tree.
+
+SafeDI is designed to make instantiating and receiving dependencies _simple_, without requiring engineers to think about abstract dependency injection (DI) concepts. That said, for those familiar with other DI systems: each `@Instantiable` type is its own [Scope](https://medium.com/@aarontharris/scope-dependency-injection-6fc25beffc9c). For those unfamiliar with DI terminology, know that each `@Instantiable` type is responsible for retaining its dependencies, and that every `@Instantiated` or `@Forwarded` dependency is available to all transitive child dependencies.  
+
+Every `@Instantiable`-decorated type must be:
+
+1. `public` or `open`
+
+2. Have a `public init(…)` or `open init(…)` that receives every injectable property
```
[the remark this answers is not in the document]
dfed: They cannot. There must be one initializer that has arguments only for only injected (Instantiable, Forwarded, or Received) properties.

Under the hood, the reason for this is that this is the initializer that SafeDI uses to initialize your Instantiable object. And since SafeDI is only aware of (and has access to) properties in the DI tree... we wouldn't be able to make a call to an initializer with additional properties.

README.md
dfed: Good call that I should make this explicit. Yeah, for the root of the dependency graph you must call `init()`. It's still true that you shouldn't call `init(...)` though... but yeah I should make this more clear.

README.md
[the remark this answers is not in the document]
dfed: Going to leave this as-is for now since:

1. I'm struggling to come up with better wording here
2. There is a FixIt if you do this wrong

README.md
dfed: 521a99a4c308e9bceef1065372647cb75fe85266

README.md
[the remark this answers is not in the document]
dfed: I need to take another pass here given 418bada – this initializer is now generated for you unless you have uninitialized properties on the type.

README.md
```diff
@@ -1,2 +1,461 @@
...
+[![Swift Package Manager compatible](https://img.shields.io/badge/SPM-✓-4BC51D.svg?style=flat)](https://github.com/apple/swift-package-manager)
+[![codecov](https://codecov.io/gh/dfed/SafeDI/branch/main/graph/badge.svg)](https://codecov.io/gh/dfed/SafeDI)
+[![License](https://img.shields.io/badge/License-MIT-blue.svg)](https://spdx.org/licenses/MIT.html)
+
+Compile-time safe dependency injection for Swift projects. SafeDI is built for engineers who want the safety and simplicity of manual dependency injection without the overhead of boilerplate code.
+
+## Features
+
+✅ Compile-time safe
+
```
gonzalonunez, line 13: Aesthetic need, I would maybe consider using regular markdown checkboxes here. If anything just to save you some space and get more content above-the-fold:

- [x] Compile-time safe
- [x] Thread safe
- [x] ...

README.md
```diff
@@ -1,2 +1,461 @@
...
+✅ Architecture independent
+
+✅ Simple integration: no DI-specific types or generics required
+
+✅ Easy testing: every type has a memberwise initializer
+
+✅ Clear error messages: never debug generated code
+
+## Using SafeDI
```
gonzalonunez, line 36: A single top-level code sample, maybe even just a condensed version of the example app's root view, would be really helpful to see before diving into the specifics behind each macro. Would really help contextualize the details below.

README.md
```diff
@@ -1,2 +1,461 @@
...
+
+The `@State` documentation reads:
+
+> Declare state as private to prevent setting it in a memberwise initializer, which can conflict with the storage management that SwiftUI provides
+
+The `@StateObject` documentation reads:
+
+> Declare state objects as private to prevent setting them from a memberwise initializer, which can conflict with the storage management that SwiftUI provides
+
+`@Instantiated`, `@Forwarded`, or `@Received` objects may be decorated with [`@ObservedObject`](https://developer.apple.com/documentation/swiftui/ObservedObject). Note that `@Instantiated` objects declared on a `View` will be re-initialized when the view is re-initialized. You can find a deep dive on SwiftUI view lifecycles [here](https://www.donnywals.com/understanding-how-and-when-swiftui-decides-to-redraw-views/).
```
gonzalonunez, line 363: This feels like an important subtlety. Maybe a code-sample or a link to a Wiki in the repo that explains in detail?

dfed: :100:. Will need to do this on a future pass, but yeah this is a tough one. The example app shows has an example of this in practice at least.

README.md
dfed: Yeah I need to do another pass on this and pull a bunch of the below out into a wiki + focus the README on "how do I get started". That'll be a future round though.

README.md
kierajmumick: While it’s important and subtle, it is a subtlety inherent in SwiftUIs architecture, with or without this DI system. 

It shouldn’t be the job of these docs to
the SwiftUI subtleties in detail. I do think the call out & link is likely enough!

dfed:
I'm going to merge this, but I am still quite open to feedback on both this documentation and the API! I'll be pushing a beta 0.1.0 tag release of this repo shortly, but since we're still in beta it's a good time to provide API feedback!

One bit of feedback I've received offline from a couple folk is that much of the content in this README would be best split out into wiki docs, and the README should be refocused on more of a Quickstart guide (with clear code examples pulled from the sample project) + linking to the rest of the wiki. This is great feedback, and I agree with it, but in the spirit of shipping I'm going to merge this as-is and take a future action item to rework this.

PR #6, opened 2024-01-07
Enable non-lexigraphic equivalence checking for Compositions

dfed:
Today, SafeDI considers `Foo & Bar` and `Bar & Foo` to be discrete types. However, the compiler does _not_ consider these types to be discrete, and therefore neither should we.

This change enables treating the above compositions as equivalent by utilizing an unordered `Set` to store the associated types of a `TypeDescription.composition`.


Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -153,7 +157,10 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
             self = .implicitlyUnwrappedOptional(typeDescription)
 
         case Self.compositionDescription:
-            let typeDescriptions = try values.decode([Self].self, forKey: .typeDescriptions)
+            let typeDescriptions = try values.decodeIfPresent(Set<Self>.self, forKey: .unorderedTypeDescriptions)
+            // For backwards compatibility for when compositions were ordered.
+            // TODO: Delete this backwards compatibility line when we make our next breaking change.
+            ?? values.decode(Set<Self>.self, forKey: .typeDescriptions)
```
dfed, line 163:
```suggestion
            ?? values.decode([Self].self, forKey: .typeDescriptions)
```

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -153,7 +157,10 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
             self = .implicitlyUnwrappedOptional(typeDescription)
 
         case Self.compositionDescription:
-            let typeDescriptions = try values.decode([Self].self, forKey: .typeDescriptions)
+            let typeDescriptions = try values.decodeIfPresent(Set<Self>.self, forKey: .unorderedTypeDescriptions)
+            // For backwards compatibility for when compositions were ordered.
```
dfed, line 161:
```suggestion
            // For backwards compatibility from when compositions were ordered.
```

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -249,6 +257,8 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
         case typeDescription
         /// The value for this key is the associated value of type [TypeDescription]
         case typeDescriptions
+        /// The value for this key is the associated value of type Set<TypeDescription>
+        case unorderedTypeDescriptions
```
dfed, line 261: I'm not convinced this is a necessary addition. While the extra specificity is nice... the underlying json representations of a `Set` and `Array` are identical (aside from ordering).

So my options are:

1. Introduce this new key
2. Update the documentation on `case typeDescriptions` to state that the value here is either an array or set.

I'm currently leaning towards the second option.

dfed: Yeah alright talked myself into it and pushed up 8afe45ea537d76a0c0aa2b80cf31f0ab06d8a6cb. Open to feedback, and could be convinced to revert.

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -27,7 +27,7 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
     /// A nested type with possible generics. e.g. Array.Element or Swift.Array<Element>
     indirect case nested(name: String, parentType: TypeDescription, generics: [TypeDescription])
     /// A composed type. e.g. Identifiable & Equatable
-    indirect case composition([TypeDescription])
+    indirect case composition(Set<TypeDescription>)
```
dfed, line 30: A downside of using a `Set` here is that we no longer know the written order of the types of the composition in the source code. We _could_ offset that by utilizing an array still, and then doing set comparisons in a manually written `==` implementation, but:

1. Doing just-in-time `Set` comparisons can get expensive when doing a lot of comparisons
2. Maintaining a manually written `==` implementation sounds... not fun

So I'm pretty bullish that this approach is good enough for our needs. Open to feedback and alternative thoughts of course.

igz: There's always `OrderedSet` from `swift-collections`, but then you'd be adding a dependency.

dfed: yup! `OrderedSet` would solve the just-in-time `Set` comparison problem. But I'd still need the manually written `==` which is a bit painful. Might be worth it though?

dfed: I have no qualms with a dependency, especially one that already supports Linux 😄 https://swiftpackageindex.com/apple/swift-collections

dfed: I'm gonna write a new type that houses both an array and a set, with its own `==` implementation, and use that.

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -71,7 +71,11 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
                 return "\(name)<\(generics.map { $0.asSource }.joined(separator: ", "))>"
             }
         case let .composition(types):
-            return types.map { $0.asSource }.joined(separator: " & ")
+            return types
+                .map { $0.asSource }
+                // Sort the result to ensure stable code generation.
+                .sorted()
```
dfed, line 77: we'll get rid of this line when we introduce the new type

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -425,6 +429,38 @@ extension ExprSyntax {
...
+    init(_ array: [Element]) {
+        self.array = array
+        set = Set(array)
+    }
+
+    let array: [Element]
+    private let set: Set<Element>
+
+    public static func == (lhs: UnorderedComparingArray, rhs: UnorderedComparingArray) -> Bool {
+        lhs.set == rhs.set
```
MrAdamBoyd, line 445: Duh, easier than I thought

dfed: Hardest part was coming up with a name for the type. I'm not convinced I did a great job there.

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -425,6 +429,38 @@ extension ExprSyntax {
     }
 }
 
+// MARK: - UnorderedComparingArray
+
+public struct UnorderedComparingArray<Element: Codable & Hashable & Sendable>: Codable, Hashable, Sendable, Collection {
```
MrAdamBoyd, line 434: What do you think about changing the name? Having a subscript for an unordered array seems a bit confusing.

MrAdamBoyd (approved):
LGTM other than one question

Sources/SafeDICore/Models/TypeDescription.swift
dfed: All sequences have subscripts, and sequences are not necessarily ordered logically (though their order is stable). It's a quirk of Foundation.

dfed: But I am super open to other names

dfed: Going with `UnorderedComparingCollection`

PR #7, opened 2024-01-09
Make it possible to rename or retype a @Received property

### LINKED ISSUE #5: It should be possible to receive a property of one name/type that is fulfilled by a different property name/type
**Goals**
* Enable a single instance to be represented by two different properties with different names/types

**Non-Goals**
* Enable SafeDI-level checks to ensure that a received property can be fulfilled by another instantiated or forwarded property

**Investigations**
It's possible to use `@Forwarded` to enable propagating a property under a new name and type, but given that you can only have one forwarded property per `@Instantiable` type, this is a limiting approach.

**Problem Statement**
Take the following code:
```
public protocol UserProvider {
    var user: User { get }
}

public protocol UserManager: UserProvider {
    func setUserName(_ userName: String)
}

@Instantiable(fulfillingAdditionalTypes: [UserManager.self])
public final class DefaultUserManager: UserManager {
    ... // implementation here
}
```

It should be possible in one place in the DI tree to instantiate a `let userManager: UserManager` and further down the DI tree receive a `let userProvider: UserProvider` whose instance is the same instance as the `userManager`'s.

**Design**
I believe the simplest path forward here is to add an optional parameter to the `@Received` macro. Something like:

```
@Received(fulfilledByDependencyNamed: "userManager")
let userProvider: UserProvider
```

SafeDI would then know to fulfill the `@Received` property with an `@Instantiated` or `@Forwarded` property further up the tree called `userManager`. SafeDI would not be able to type check to ensure that the `userManager` property actually conforms to `UserProvider`, but if it can't then the generated code would fail to compile in a way that would make the issue fairly obvious.

Note that this design only applies to cases where the received property is of a type the desired dependency _already directly conforms to_. If `UserManager` did not inherit from `UserProvider`, we'd be in trouble. We can add more use-at-your-own-risk flexibility to SafeDI's macros by adding yet another optional parameter to the `@Recevied` macro:

```
@Received(fulfilledByDependencyNamed: "userManager", forceCast: true)
let userProvider: UserProvider
```

This would cause the generated code to force cast the `userManager as! UserProvider`. I don't expect we'll introduce the `forceCast` option in the near term, but I outline it here to document how we could solve such a problem down the line.

dfed:
Resolves #5. Feel free to provide API or code-level comments when reviewing!


README.md
```diff
@@ -179,10 +180,6 @@ public struct ParentView: View {
         }
     }
 
-    public init(childViewBuilder: Instantiator<some View>) {
```
dfed, line 182: I removed these `init` methods from the README that are auto-generated now.

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -29,23 +29,29 @@ actor ScopeGenerator {
         propertiesToGenerate: [ScopeGenerator],
         receivedProperties: Set<Property>
     ) {
-        self.instantiable = instantiable
+        if let property {
+            scopeData = .property(
```
dfed, line 33: I've now created a `ScopeData` enum to encapsulate the different kinds of scopes that exist that we need to generate for. Scroll down to find the definition.

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -67,13 +87,28 @@ actor ScopeGenerator {
...
-                if let property {
+                switch scopeData {
+                case let .root(instantiable):
+                    let argumentList = try instantiable.generateArgumentList()
+                    if instantiable.dependencies.isEmpty {
+                        // Nothing to do here! We already have an empty initializer.
+                        return ""
+                    } else {
+                        return """
+                            extension \(instantiable.concreteInstantiableType.asSource) {
```
dfed, line 103: this code-gen moved from below without changes

Sources/SafeDICore/Models/Dependency.swift
```diff
@@ -24,15 +24,44 @@ import SwiftSyntax
...
     public let property: Property
+    /// The source of the dependency within the dependency tree.
     public let source: Source
+    /// The name of the property that will be used to fulfill this property.
+    public let fulfillingPropertyName: String?
+    /// The type description of the type that will be used to fulfill this property.
+    /// This type must be the same as or a parent type of `property.typeDescription`.
     public let fulfillingTypeDescription: TypeDescription?
-
-    public var asInstantiatedType: TypeDescription {
```
dfed, line 33: I removed just-in-time calculated types to avoid recalculating/reallocating the same object/value over and over. No actual changes here.

README.md
```diff
@@ -76,6 +76,7 @@ public final class UserService {
     // MARK: Initialization
 
     /// Public, memberwise initializer that takes each injected property.
+    /// This initializer can be auto-generated by SafeDI if it isn‘t manually written.
```
dfed, line 79: Seemed like a good thing to add, though it is unrelated to the main thrust of this PR.

Sources/SafeDICore/Models/Dependency.swift
```diff
@@ -24,15 +24,44 @@ import SwiftSyntax
...
+        self.fulfillingTypeDescription = fulfillingTypeDescription
+
+        if let fulfillingPropertyName, let fulfillingTypeDescription {
+            fulfillingProperty = Property(
+                label: fulfillingPropertyName,
+                typeDescription: fulfillingTypeDescription)
+        } else {
+            fulfillingProperty = nil
+        }
+        asInstantiatedType = (fulfillingTypeDescription ?? property.typeDescription).asInstantiatedType
```
dfed, line 47: this code was moved up from below without changes

README.md
MrAdamBoyd: Shouldn't it be this? "This initializer **will** be auto-generated..."

MrAdamBoyd (approved):
LGTM aside from one small comment

README.md
dfed: It is only auto-generated if certain conditions are met. Specifically, we generate the initializer if there is no uninitialized, non-optional SafeDI property on the type. Otherwise we can't generate an initializer that will compile.

PR #8, opened 2024-01-10
Use generated Codable implementation for TypeDescription

dfed:
Back when we first wrote `TypeDescription` [a few years back](https://github.com/fdiaz/SwiftInspector/pull/69), Swift was not capable of generating the `Codable` implementation for indirect enums with associated values. Now it is.

Instead of continuing to maintain and test a manual `Codable` implementation, let's rely on Apple's implementation. A nice bonus is that Apple's implementation is more compact than our implementation, which will likely improve encoding/decoding speed for folk utilizing persistence.

This is a breaking change.


dfed:
Merging this, but will wait a bit before releasing since there are going to be a few breaking changes landing. If folk have feedback post-merge please leave a comment!

PR #12, opened 2024-01-13
Add 'nonisolated' to generated initializers

dfed:
`struct` and `class` types may be `@MainActor`-decorated, at which point we need our generated initializer to be `nonisolated`


MrAdamBoyd (approved):
Easy

PR #13, opened 2024-01-13
Improve test coverage

dfed:
Exercises more of the SafeDI library via tests, and discovers/fixes an issue where an source empty file list in the CSV would cause an error.

There's also a little code cleanup along the way here.

### DIFF
--- a/Sources/SafeDICore/Visitors/InstantiableVisitor.swift
+++ b/Sources/SafeDICore/Visitors/InstantiableVisitor.swift
@@ -57,9 +57,8 @@ public final class InstantiableVisitor: SyntaxVisitor {
             let patterns = node
                 .bindings
                 .filter {
-                    $0.initializer == nil
-                    && $0.accessorBlock == nil
-                    && !$0.isOptionalAndUninitialized
+                    $0.accessorBlock == nil
+                    && !$0.isOptionalOrInitialized
                 }
                 .map(\.pattern)
             uninitializedNonOptionalPropertyNames += patterns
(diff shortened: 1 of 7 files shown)


Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -57,9 +57,8 @@ public final class InstantiableVisitor: SyntaxVisitor {
             let patterns = node
                 .bindings
                 .filter {
-                    $0.initializer == nil
```
dfed, line 60: this check was duplicative with `isOptionalOrInitialized` below

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -57,9 +57,8 @@ public final class InstantiableVisitor: SyntaxVisitor {
...
-                    && $0.accessorBlock == nil
-                    && !$0.isOptionalAndUninitialized
```
dfed, line 62: `isOptionalAndUninitialized` was named incorrectly. It was really checking whether something was optional _or_ uninitialized

Tests/SafeDICoreTests/TypeDescriptionTests.swift
```diff
@@ -542,9 +542,9 @@ final class TypeDescriptionTests: XCTestCase {
         XCTAssertEqual(typeDescription.asSource, "() async throws -> ()")
     }
 
-    func test_asSource_whenDescribingAnUnknownCase_returnsTheProvidedStringWithWhitespaceStripped() {
```
dfed, line 545: turns out we're not stripping leading whitespace. But that's honestly fine since we have no valid, compiling cases that result in an `.unknown` type.

This test wasn't previously testing the `.unknown` case, but rather the `.simple` case.

dfed:
Merging this without review since we aren't changing API – we're just improving coverage.

PR #14, opened 2024-01-15
Ensure platform support in CI

dfed:
`path()`, `URL.temporaryDirectory`, and `appending(path:)` have limited OS availability, and our CI only tests against latest operating systems (due to limitations with GitHub Actions). Swift package index's test suite was able to catch an issue with our use of `path()` ([iOS](https://swiftpackageindex.com/builds/F78DED19-7319-48DE-988C-858A00241AC9), [watchOS](https://swiftpackageindex.com/builds/E84BC0F8-0D4B-48C2-83B1-1B6B9812DD29), and [tvOS](https://swiftpackageindex.com/builds/A333E309-D9AA-4562-ADEE-C7D5DF8A2C60)), and we copy their build config here.

Thankfully these availability issues only occurred in the test suite rather than the production suite, so `0.2.1` has the same version support as `0.2.0` despite the Swift Package Index build failures linked above.

I also moved to utilizing `relativePath` instead of `path()` and `path` since `relativePath` does not have availability issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   spm-15:
     name: Build and Test on Xcode 15
-    runs-on: macos-13-arm64
+    runs-on: macos-latest-xlarge
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
@@ -25,7 +25,7 @@ jobs:
 
   xcodebuild-15:
     name: Build with xcodebuild on Xcode 15
-    runs-on: macOS-13-arm64
+    runs-on: macos-latest-xlarge
     strategy:
       matrix:
         platforms: [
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   spm-15:
     name: Build and Test on Xcode 15
-    runs-on: macos-13
+    runs-on: macos-13-arm64
```
dfed, line 12: this change isn't required but I'm hoping arm64 machines will execute faster

.github/workflows/ci.yml
```diff
@@ -23,9 +23,29 @@ jobs:
...
+  xcodebuild-15:
+    name: Build with xcodebuild on Xcode 15
+    runs-on: macOS-13-arm64
+    strategy:
+      matrix:
+        platforms: [
+          'generic/platform=ios',
+          'platform=macos,arch=arm64',
+          'generic/platform=tvos',
+          'generic/platform=watchos',
```
dfed, line 35: this test matrix was stolen from swift package index's builds

PR #17, opened 2024-01-16
Run CI on visionOS

dfed:
As title


dfed:
Woof. We can't do this until GitHub enables Apple Silicon Macs for open source projects.

PR #20, opened 2024-01-18
Throw error when ForwardingInstantiator is used when Instantiator should be used

dfed:
I took a type that had a `@Forwarded` property and made the property `@Received`, didn't replace the `ForwardingInstantiator` with an `Instantiator`, and then I ended up with a build error in the codegen. That's bad! I should have gotten a plugin-time error like the one we're introducing here.


MrAdamBoyd (approved):
Very nice, easy

PR #21, opened 2024-01-18
Always use local instantiated or aliased properties over inherited properties

dfed:
When a property is `@Instantiated` both above the current scope and within the current scope, we should *always* rely on the property that is instantiated in the current scope – previously it was a bit of a crapshoot and we could end up with code that couldn't compile.

This hard-and-fast "always use the local version" rule miiiight be problematic in some weird use cases, but at a high level if you need to mix using the outside instance and the inner instance of the same type, just pick different names for your variables rather than overriding an existing instance with the same name.


Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -1588,6 +1588,75 @@ final class SafeDIToolTests: XCTestCase {
         )
     }
 
+    func test_run_writesConvenienceExtensionOnRootOfTree_whenRootInstantiatesPropertiesWithSingleTreeThatInstantiateTheSamePropertyAtMultipleLevels() async throws {
```
dfed, line 1591: this test failed before I made the above change

MrAdamBoyd (approved):
Glad to be the rubber duck 🦆

PR #22, opened 2024-01-19
Ensure equivalent UnorderedComparingCollections have equivalent hash values

dfed:
`UnorderedComparingCollections`'s `hashValue` is incorrect which leads to unexpected behavior: a `hash(into:)` function should inspect the same elements that our `==` function does, which it does not today.


MrAdamBoyd (approved):
LGTM

PR #23, opened 2024-01-19
Enable aliased received properties to be received again further down the chain

dfed:
Once we renamed a property, we were not able to receive that renamed property further down the chain.


Sources/SafeDICore/Models/Scope.swift
```diff
@@ -55,10 +55,10 @@ final class Scope {
             .compactMap {
                 switch $0.source {
                 case .received:
-                    return $0.fulfillingProperty ?? $0.property
+                    $0.fulfillingProperty ?? $0.property
                 case .forwarded,
                         .instantiated:
-                    return nil
+                    nil
```
dfed, line 61: this is a style change. not related to the underlying fix here

Sources/SafeDICore/Models/Scope.swift
```diff
@@ -81,7 +81,12 @@ final class Scope {
                 instantiableStack
                     .flatMap(\.dependencies)
                     .filter {
-                        $0.source != .received
+                        (
+                            // If the source is not received, the property has been made available.
+                            $0.source != .received
+                            // If the dependency has a fulfilling property, the property has been aliased.
+                            || $0.fulfillingProperty != nil
```
dfed, line 88: This matches similar code that we have over in the ScopeGenerator: https://github.com/dfed/SafeDI/blob/e1d20ca533ffc569ad910bcf84cfbc11df68a49c/Sources/SafeDICore/Generators/ScopeGenerator.swift#L55-L58

dfed: To be honest the more I think about this (and the PR built on top of this one)... the more I think the abstraction here is wrong. This change works and is mergeable, however I may take a bigger pass at cleaning this up before the next PR makes it out of draft.

dfed: Yeah alright pushed up something _way_ better in #24.

dfed:
Closing in favor of #26

PR #24, opened 2024-01-19
Enable aliasing a property that is forwarded in at the same level

dfed:
Fixes an issue where we could rename properties that were forwarded in above us, but not at the same level.

Due to our insight in #23 https://github.com/dfed/SafeDI/pull/23#discussion_r1458189333 (the PR on which this depends), I have rewritten `Dependency.Source` to have a separate case for an alias. By adding associated values to the enum, we were able to remove multiple optional properties and hard-to-keep-track-of tribal knowledge in the codebase.

Because we have updated the `Dependency.Source` which `SafeDITool` may persist, **this PR is a breaking change** and we will need to roll to a 0.3.0 when this merges.

On top of that, this is a pretty meaty change. **I recommend reviewing this change in side-by-side view and turning off whitespace diffs** to best understand what we were accomplishing before vs after.


.github/workflows/ci.yml
```diff
@@ -68,13 +68,10 @@ jobs:
   linux:
     name: Build and Test on Linux
     runs-on: ubuntu-latest
+    container: swift:5.9
```
dfed, line 71: following the advice [here](https://github.com/swift-actions/setup-swift/issues/591#issuecomment-1685710678) to avoid [this kind of CI failure](https://github.com/dfed/SafeDI/actions/runs/7581297537/job/20648662318)

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -256,9 +263,17 @@ public final class DependencyTreeGenerator {
...
+                    .instantiable
+                    .dependencies
+                    .filter(\.isForwarded)
+                    .map(\.property)
+            )
             for receivedProperty in scope.receivedProperties {
                 let parentContainsProperty = receivableProperties.contains(receivedProperty)
-                if !parentContainsProperty {
+                let propertyIsForwarded = forwardedProperties.contains(receivedProperty)
+                if !parentContainsProperty && !propertyIsForwarded {
```
dfed, line 276: this is the crux of the change that makes the new tests pass. We were already inspecting properties that were forwarded by a higher-up scope (they're in `receivedProperties`), but we weren't inspecting the properties in _out_ scope.

Sources/SafeDICore/Extensions/AttributeListSyntaxExtensions.swift
```diff
@@ -24,12 +24,7 @@ extension AttributeListSyntax {
 
     public var instantiableMacro: AttributeSyntax? {
         guard let attribute = first(where: { element in
-            switch element {
-            case let .attribute(attribute):
-                return IdentifierTypeSyntax(attribute.attributeName)?.name.text == InstantiableVisitor.macroName
-            case .ifConfigDecl:
-                return false
-            }
+            element.instantiableMacro != nil
```
dfed, line 27: I simplified our macro access while I was adding associated types to the `Dependency.Source` enum in order to share this code.

Sources/SafeDICore/Errors/FixableInstantiableError.swift
```diff
@@ -35,7 +35,7 @@ public enum FixableInstantiableError: DiagnosticError {
     public var description: String {
         switch self {
         case .dependencyHasTooManyAttributes:
-            "Dependency can have at most one of @\(Dependency.Source.instantiated), @\(Dependency.Source.received), or @\(Dependency.Source.forwarded) attached macro"
```
dfed, line 38: `Dependency.Source` is no longer a `String`-backed enum, so we need to use constants for these raw values now.

dfed:
Closing in favor of #26

PR #26, opened 2024-01-19
Refactor property validation and simplify generation

dfed:
**This PR is best reviewed commit by commit, in split view, with whitespace ignored**

There's a lot going on here, but the high level is that we are refactoring how we store and generate dependencies to remove a ton of potential issues.

We had a ton of little bugs because we were:
1. Storing associated information as optional properties away from associated enum cases
2. Validating our dependency tree twice – once in the dependency tree generator and once during code generation

We fix both of these architectural issues in this PR (in separate commits), as well as a few other minor bugs that I found along the way.

Because we have updated the `Dependency.Source` – which `SafeDITool` may persist – **this PR is a breaking change** and we will need to roll to a 0.3.0 when this merges.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,13 +68,10 @@ jobs:
   linux:
     name: Build and Test on Linux
     runs-on: ubuntu-latest
+    container: swift:5.9
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
-      - name: Setup Swift Environment
-        uses: swift-actions/setup-swift@v1
-        with:
-          swift-version: 5.9
       - name: Build and Test Framework
         run: swift test -c release --enable-code-coverage -Xswiftc -enable-testing
       - name: Prepare Coverage Reports
--- a/Sources/SafeDICore/Errors/FixableInstantiableError.swift
+++ b/Sources/SafeDICore/Errors/FixableInstantiableError.swift
@@ -35,7 +35,7 @@ public enum FixableInstantiableError: DiagnosticError {
     public var description: String {
         switch self {
         case .dependencyHasTooManyAttributes:
-            "Dependency can have at most one of @\(Dependency.Source.instantiated), @\(Dependency.Source.received), or @\(Dependency.Source.forwarded) attached macro"
+            "Dependency can have at most one of @\(Dependency.Source.instantiatedRawValue), @\(Dependency.Source.receivedRawValue), or @\(Dependency.Source.forwardedRawValue) attached macro"
         case .dependencyHasInitializer:
             "Dependency must not have hand-written initializer"
         case .missingPublicOrOpenAttribute:
--- a/Sources/SafeDICore/Extensions/AttributeListSyntaxExtensions.swift
+++ b/Sources/SafeDICore/Extensions/AttributeListSyntaxExtensions.swift
@@ -24,12 +24,7 @@ extension AttributeListSyntax {
 
     public var instantiableMacro: AttributeSyntax? {
         guard let attribute = first(where: { element in
-            switch element {
-            case let .attribute(attribute):
-                return IdentifierTypeSyntax(attribute.attributeName)?.name.text == InstantiableVisitor.macroName
-            case .ifConfigDecl:
-                return false
-            }
+            element.instantiableMacro != nil
         }) else {
             return nil
         }
@@ -38,12 +33,7 @@ extension AttributeListSyntax {
 
     public var instantiatedMacro: AttributeSyntax? {
         guard let attribute = first(where: { element in
-            switch element {
-            case let .attribute(attribute):
-                return IdentifierTypeSyntax(attribute.attributeName)?.name.text == Dependency.Source.instantiated.rawValue
-            case .ifConfigDecl:
-                return false
-            }
+            element.instantiatedMacro != nil
         }) else {
             return nil
         }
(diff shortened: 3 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -68,13 +68,10 @@ jobs:
   linux:
     name: Build and Test on Linux
     runs-on: ubuntu-latest
+    container: swift:5.9
```
dfed, line 71: following the advice [here](https://github.com/swift-actions/setup-swift/issues/591#issuecomment-1685710678) to avoid [this kind of CI failure](https://github.com/dfed/SafeDI/actions/runs/7581297537/job/20648662318)

Sources/SafeDICore/Extensions/AttributeListSyntaxExtensions.swift
```diff
@@ -24,12 +24,7 @@ extension AttributeListSyntax {
 
     public var instantiableMacro: AttributeSyntax? {
         guard let attribute = first(where: { element in
-            switch element {
-            case let .attribute(attribute):
-                return IdentifierTypeSyntax(attribute.attributeName)?.name.text == InstantiableVisitor.macroName
-            case .ifConfigDecl:
-                return false
-            }
+            element.instantiableMacro != nil
```
dfed, line 27: I simplified our macro access while I was adding associated types to the `Dependency.Source` enum in order to share this code.

Sources/SafeDICore/Errors/FixableInstantiableError.swift
```diff
@@ -35,7 +35,7 @@ public enum FixableInstantiableError: DiagnosticError {
     public var description: String {
         switch self {
         case .dependencyHasTooManyAttributes:
-            "Dependency can have at most one of @\(Dependency.Source.instantiated), @\(Dependency.Source.received), or @\(Dependency.Source.forwarded) attached macro"
+            "Dependency can have at most one of @\(Dependency.Source.instantiatedRawValue), @\(Dependency.Source.receivedRawValue), or @\(Dependency.Source.forwardedRawValue) attached macro"
```
dfed, line 38: `Dependency.Source` is no longer a `String`-backed enum, so we need to use constants for these raw values now.

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -2521,7 +3172,7 @@ final class SafeDIToolTests: XCTestCase {
         }
     }
 
-    func test_run_onCodeWithUnfulfillableRenamedReceivedPropertyType_throwsError() async {
```
dfed, line 2524: I made the tests refer to `Aliased` rather than `Renamed` for consistency. We still talk about renaming in the README, but that's fine.

I'm also open to feedback that the `case` should be `case renamed` rather than `case alias`

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -2350,6 +2350,600 @@ final class SafeDIToolTests: XCTestCase {
...
+                        return LoggedInViewController(userManager: userManager, userVendor: userVendor, profileViewControllerBuilder: profileViewControllerBuilder)
+                    }
+                    self.init(authService: authService, networkService: networkService, loggedInViewControllerBuilder: loggedInViewControllerBuilder)
+                }
+            }
+            """
+        )
+    }
+
+    func test_run_writesConvenienceExtensionOnRootOfTree_whenReceivedPropertyIsAliasedWasALevelBelowWhereItWasForwarded() async throws {
```
dfed, line 2805: This test did already succeed, but it still seemed worth having this coverage.

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -216,33 +216,40 @@ public final class DependencyTreeGenerator {
...
+                    continue
                 }
-                additionalPropertiesToInstantiate.append(Scope.PropertyToInstantiate(
-                    property: instantiatedDependency.property,
-                    instantiable: instantiable,
-                    scope: instantiatedScope,
-                    type: type
-                ))
             }
             scope.propertiesToInstantiate.append(contentsOf: additionalPropertiesToInstantiate)
```
dfed: `propertiesToInstantiate` might not be the best name. `propertiesToGenerate`?

Sources/SafeDICore/Models/Scope.swift
```diff
@@ -81,38 +99,40 @@ final class Scope {
...
+                                .forwarded,
+                            // This property has been re-injected into the dependency tree under a new alias.
+                                .aliased:
+                            return !propertyStack.contains($0.property) && $0.property != property
+                        case .received:
+                            return false
+                        }
                     }
                     .map(\.property)
             ).subtracting(
```
dfed: do we need `receivedProperties` calculated for both cases? could simplify maybe

dfed: ah we got rid of this anyways

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
dfed: indeed. 346e1daf6bda1452f11c6e78b2cf836a00c98d6e

MrAdamBoyd (approved):
Thank you for the walkthrough, approved

PR #28, opened 2024-01-20
Improve custom collection usage, performance, and dependency ordering fixes

dfed:
Improves how we utilize the `List` introduced in #27. Simplifications, performance improvements, and a few more bug fixes (see tests)


Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -249,33 +249,29 @@ actor ScopeGenerator {
         let propertiesToGenerate = Set(propertiesToGenerate.compactMap(\.property))
         for propertyToGenerateNode in orderedPropertiesToGenerate {
             let hasDependenciesGeneratedByCurrentScope = !propertyToGenerateNode
-                .value
```
dfed, line 252: we no longer need this because we've implemented `@dynamicMemberLookup` on `List`

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -249,33 +249,29 @@ actor ScopeGenerator {
...
                 .requiredReceivedProperties
                 .isDisjoint(with: propertiesToGenerate)
             guard hasDependenciesGeneratedByCurrentScope else {
                 // This property does not have received dependencies generated by this scope, therefore its ordering is irrelevant.
                 continue
             }
-            var lastDependency: List<ScopeGenerator>?
-            for nextPropertyToGenerate in propertyToGenerateNode.dropFirst() {
+            for futurePropertyToGenerate in propertyToGenerateNode.dropFirst().reversed() {
```
dfed, line 258: reversing allows us to stop iterating as soon as we hit a dependency since we know we're at the last one.

Sources/SafeDICore/Models/List.swift
```diff
@@ -26,6 +27,8 @@ public final class List<Element>: Sequence {
         self.value = value
         self.previous = previous
         self.next = next
+        previous?.next = self
+        next?.previous = self
```
dfed, line 31: now our initializer makes suer the `previous` and `next` nodes aren't a lie.

a little odd to have an initializer side effect on other instances... but without it this initializer can leave us in a pretty bad place.

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -1657,6 +1657,77 @@ final class SafeDIToolTests: XCTestCase {
         )
     }
 
+    func test_run_writesConvenienceExtensionOnRootOfTree_whenRootInstantiatesPropertiesWithSingleTreeThatInstantiatesAReceivedPropertyBelowWhereItIsReceived() async throws {
```
dfed, line 1660: this test did not pass prior to the changes in this PR.

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -45,7 +45,6 @@ actor ScopeGenerator {
         }
         self.property = property
         self.propertiesToGenerate = propertiesToGenerate
-        forwardedProperties = scopeData.forwardedProperties
```
dfed, line 48: we no longer need `forwardedProperties`

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -78,8 +83,7 @@ actor ScopeGenerator {
                         return fulfillingProperty
                     }
                 }
-        ).union(propertiesToGenerate.flatMap(\.requiredReceivedProperties))
-            .subtracting(propertiesMadeAvailableByChildren)
```
dfed, line 82: subtracting the `propertiesMadeAvailableByChildren` at the end was _wrong_, since it erased any properties we needed at this level! I re-ordered the operations above to ensure we always include the properties we need

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -45,27 +45,19 @@ actor ScopeGenerator {
         }
         self.property = property
         self.propertiesToGenerate = propertiesToGenerate
-        forwardedProperties = scopeData.forwardedProperties
-        propertiesMadeAvailableByChildren = Set(
```
dfed, line 49: calculating this as a union recursive list created issues when we re-instantiate the same property multiple times in the tree. We now inspect each tree individually (depth first) to avoid this trap.

PR #30, opened 2024-01-20
Improve handling of forwarded properties

dfed:
* Ensure that forwarded properties do not affect dependency ordering – this issue did not lead to non-compiling codegen, but our code-gen is more correct now.
* Throw error when trying to instantiate a property that has forwarded properties outside of a ForwardingInstantiator


Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -1716,6 +1716,114 @@ final class SafeDIToolTests: XCTestCase {
         )
     }
 
+    func test_run_writesConvenienceExtensionOnRootOfTree_whenRootInstantiatesAPropertyForwardedByAChild_doesNotRequirePuttingInstantiatedPropertyBeforeForwardingInstantiator() async throws {
```
dfed, line 1719: prior to the above changes, this test output compiling code, but this code was ordered sub-optimally.

PR #33, opened 2024-01-21
Use insertion sort on dependencies to generate to reduce time and space complexity of sort

dfed:
Yet another performance improvement. We're now using an insertion sort algorithm that allows us to iterate over `propertiesToGenerate` a single time, and doing a single set of O(<N) operations per property.

Our algorithm requires significantly less space complexity now since we are not creating subsequences per insertion.


Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -236,37 +236,39 @@ actor ScopeGenerator {
...
+            return propertiesToGenerate
+        }
+        for propertyToGenerate in propertiesWithDependencyOnThisScope {
             if let lastDependencyToGenerate = propertyToGenerate
                 // Ignore our own property.
                 .dropFirst()
-                // Reverse the list to find the last property we depend on first.
-                .reversed()
-                .first(where: { futurePropertyToGenerate in
+                .lazy
```
dfed, line 252: `lazy` enables us to `filter` without allocating a new intermediary array.

dfed:
Shoot. Reverting this. There's an issue with the insertion-sort algorithm. Specifically, I think we're inserting something with dependencies and then inserting one of its dependencies into the wrong place later.

I'm going to waive off further optimizations here. This code isn't even the thing showing up in timer profiles, and what we have works well.

A Topological Sort is probably the best algorithm we have, but the memory complexity is significantly worse than what we're doing here.

PR #34, opened 2024-01-22
Detect cycles in received properties

dfed:
We weren't detecting cycles in received properties, and in our previous linked list insertion algorithm we'd end up in an infinite loop.

Now we properly detect received cycles. To get there, we migrated to utilizing a Topological Sort. This is a less space-efficient sort, but to detect cycles efficiently we need it.


Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -236,42 +262,90 @@ actor ScopeGenerator {
...
+
+                            if newDependenciesCount == 0 {
+                                nextPropertiesToGenerate.append(dependantScope)
+                            }
+                        }
+                    }
+                }
+            }
+
+            // Step 3: Check for cycle
```
dfed, line 318: we were detecting cycles in instantiated properties but not received properties!!

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -3085,6 +3085,338 @@ final class SafeDIToolTests: XCTestCase {
         )
     }
 
+    func test_run_writesConvenienceExtensionOnRootOfTree_whenRootHasLotsOfDependenciesThatDependOnOneAnother() async throws {
```
dfed, line 3088: complex dependency test that should help us avoid introducing real problems.

PR #41, opened 2024-01-24
Generated initializers should decorate closure types with @escaping

### LINKED ISSUE #39: Generated initializers should decorate closure types with `@escaping`
**Steps to reproduce**
Write the following type:
```
@Instantiable
public struct ClosureContainer {
    @Forwarded
    let closure: () -> Void
}
```

**Expected behavior**
SafeDI would generate an initializer:
```
public init(closure: @escaping () -> Void) {
    self.closure = closure
}
```

**Actual behavior**
SafeDI generates an initializer that does not compile because the closure is not escaping:
```
public init(closure: () -> Void) {
    self.closure = closure
}
```

Worse yet, SafeDI generates this initializer _even if you have written the following initializer_:
```
public init(closure: @escaping () -> Void) {
    self.closure = closure
}
```

**This means it is not possible to have a `@Forwarded` closure work with SafeDI today**

Offending code lives in `Initializer.generateRequiredInitializer`. A PR addressing this issue should add the following test:
```
func test_declaration_generatesRequiredInitializerWithClosureDependency() {
    assertMacro {
        """
        @Instantiable
        public struct ExampleService {
            @Forwarded
            let closure: () -> Void
        }
        """
    } expansion: {
        """
        public struct ExampleService {
            let closure: () -> Void

            // A generated initializer that has one argument per SafeDI-injected property.
            // Because this initializer is generated by a Swift Macro, it can not be used by other Swift Macros.
            // As a result, this initializer can not be used within a #Preview macro closure.
            // This initializer is only generated because you have not written this macro yourself.
            // Copy/pasting this generated initializer into your code will enable this initializer to be used within other Swift Macros.
            nonisolated public init(closure: @escaping () -> Void) {
                self.closure = closure
            }
        }
        """
    }
}
```

dfed:
Resolves #39


Sources/SafeDICore/Models/Initializer.swift
```diff
@@ -177,7 +177,7 @@ public struct Initializer: Codable, Hashable {
         try arguments.reduce(into: [(dependency: Dependency, argument: Argument)]()) { partialResult, argument in
             guard let dependency = dependencies.first(where: {
                 $0.property.label == argument.innerLabel
-                && $0.property.typeDescription == argument.typeDescription
+                && $0.property.typeDescription.isEqualToFunctionArgument(argument.typeDescription)
```
MrAdamBoyd, line 180: Any reason why you don't override `==`?

MrAdamBoyd (approved):
Single question, otherwise LGTM

Sources/SafeDICore/Models/Initializer.swift
dfed: because these things are _not_ equal. They're only equal in this specific property <-> function argument comparison context.

MrAdamBoyd: Ahhh gotcha

PR #42, opened 2024-01-25
Keep track of Void TypeDescription, and unwrap single-element Tuple types

dfed:
This PR does two things:
1. Enables `TypeDescription` to keep track of `Void` in both of its forms (`()` and `Void`) and compare them as equivalent
2. Makes `TypeDescription` unwrap single-element tuples as their underlying type. Single-element tuples are functionally equivalent in Swift to the underlying type, so we unwrap them when parsing rather than having to handle their equivalence downstream.


NickEntin (approved):

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -263,23 +301,34 @@ extension TypeSyntax {
...
                     label: $0.secondName?.text ?? $0.firstName?.text,
                     typeDescription: $0.type.typeDescription
                 )
-            })
+            }
+            if elements.isEmpty {
+                return .void(.tuple)
+            } else if elements.count == 1 {
+                // A type wrapped in a tuple is equivalent to the underlying type.
+                // To avoid handling complex comparisons later, just strip the type.
```
MrAdamBoyd, line 327: :D

MrAdamBoyd (approved):
I like this

PR #43, opened 2024-01-25
Enable a dependency root to receive a renamed @Instantiated property

### LINKED ISSUE #19: Enable a dependency root to @Received(fulfilledByDependencyNamed:ofType:) an @Instantiated property
**Goals**
* Share a single instance as multiple properties with different names or types from a root object
* Continue to ensure at plugin runtime that received properties are instantiated prior to being received

**Non-Goals**
* Ensure at plugin runtime that the new type of the received property can be fulfilled by the desired dependency (i.e. do not change how `@Received(fulfilledByDependencyNamed:ofType:)` works today). 

**Investigations**
This can be accomplished today by utilizing a `@Instantiated` extension whose `instantiate()` method always returns the same instance for multiple types, but this is a limiting way to accomplish the above goals since it ensures that we'll only ever have a single instance, which might not be what we want.

Right now, the only thing preventing roots from having an aliased received property is our own code that identifies roots.

**Design**
The design here would enable a root object to declare a `@Received(fulfilledByDependencyNamed:ofType:)` for a property that is `@Instantiated` on the same type. This will require updating `DependencyTreeGenerator`'s `lazy var possibleRootInstantiableTypes` implementation, or alternatively updating the name and implementation of `fileprivate var areAllInstantiated: Bool`.

One thing to consider is that each of a possible root type's `@Received(fulfilledByDependencyNamed:ofType:)` properties _must refer to `@Instantiated` property_ declared on that same `Instantiable` struct. Doing this check performantly will require creating a `Set` of an `Instantiable`'s dependencies. We can likely create this set just-in-time and throw it away, since we don't have another use for this `Set` just yet, and other code currently relies on the ordering of `public let dependencies: [Dependency]` on `Instantiable`.

The PR that implements this issue should have the following tests added to `SafeDIToolTests`:

##### A test to ensure that we identify a root with an aliased property
```
func test_run_writesConvenienceExtensionOnRootOfTree_whenRootHasReceivedAliasOfInstantiable() async throws {
    let output = try await executeSystemUnderTest(
        swiftFileContent: [
            """
            @Instantiable
            public struct Root {
                @Instantiated
                private let defaultUserService: DefaultUserService

                @Received(fulfilledByDependencyNamed: "defaultUserService", ofType: DefaultUserService.self)
                private let userService: any UserService
            }
            """,
            """
            import Foundation

            public protocol UserService {
                var userName: String? { get set }
            }

            @Instantiable(fulfillingAdditionalTypes: [UserService.self])
            public final class DefaultUserService: UserService {
                public init() {}

                public var userName: String?
            }
            """,
        ],
        buildDependencyTreeOutput: true
    )

    XCTAssertEqual(
        try XCTUnwrap(output.dependencyTree),
        """
        // This file was generated by the SafeDIGenerateDependencyTree build tool plugin.
        // Any modifications made to this file will be overwritten on subsequent builds.
        // Please refrain from editing this file directly.

        #if canImport(Foundation)
        import Foundation
        #endif

        extension Root {
            public init() {
                let defaultUserService: DefaultUserService = DefaultUserService()
                let userService: any UserService = defaultUserService
                self.init(defaultUserService: defaultUserService, userService: userService)
            }
        }
        """
    )
}
```

##### A test to ensure that we do not identify an instantiable with an aliased property as a root if that aliased property is not found on this type
```
func test_run_successfullyGeneratesOutputFileWhenNoRootFound() async throws {
    let output = try await executeSystemUnderTest(
        swiftFileContent: [
            """
            @Instantiable
            public struct NotRoot {
                @Instantiated
                private let defaultUserService: DefaultUserService

                // This received property's alias is improperly configured, meaning that this type is not a root.
                @Received(fulfilledByDependencyNamed: "userService", ofType: DefaultUserService.self)
                private let userService: any UserService
            }
            """,
            """
            import Foundation

            public protocol UserService {
                var userName: String? { get set }
            }

            @Instantiable(fulfillingAdditionalTypes: [UserService.self])
            public final class DefaultUserService: UserService {
                public init() {}

                public var userName: String?
            }
            """,
        ],
        buildDependencyTreeOutput: true
    )

    XCTAssertEqual(
        try XCTUnwrap(output.dependencyTree),
        """
        // This file was generated by the SafeDIGenerateDependencyTree build tool plugin.
        // Any modifications made to this file will be overwritten on subsequent builds.
        // Please refrain from editing this file directly.

        // No root @Instantiable-decorated types found, or root types already had a `public init()` method.
        """
    )
}
```

This PR should also include an update to the README to mention this capability! This may be the toughest part of the task.

dfed:
Resolves #19


README.md
```diff
@@ -247,7 +247,7 @@ public final class UpdateUserService {
 
 #### Renaming and retyping dependencies
 
-It is possible to rename or retype a dependency that is `@Instantiated` or `@Forwarded` by an object higher up in the dependency tree with the `@Received` macro. Renamed or retyped dependencies are able to be received with their new name and type by objects instantiated further down the dependency tree.
+It is possible to rename or retype a dependency that is `@Instantiated` or `@Forwarded` by an object higher up in the dependency tree with the `@Received(fulfilledByDependencyNamed:ofType:)` macro. Renamed or retyped dependencies are able to be received with their new name and type by objects instantiated further down the dependency tree.
```
dfed, line 250: not directly related to the thrust of this PR, but I figured referring to the renaming macro consistently in the README was a good idea

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -296,11 +296,17 @@ public final class DependencyTreeGenerator {
...
                 let propertyIsCreatedAtThisScope = createdProperties.contains(receivedProperty)
                 if !parentContainsProperty && !propertyIsCreatedAtThisScope {
-                    unfulfillableProperties.insert(.init(
-                        property: receivedProperty,
-                        instantiable: scope.instantiable,
-                        parentStack: instantiables.elements)
-                    )
+                    if instantiables.elements.isEmpty {
+                        // This property's scope is not a real root instantiable! Remove it from the list.
+                        rootInstantiableTypes.remove(scope.instantiable.concreteInstantiableType)
```
dfed, line 301: a little bit of hax, but when we're validating properties and we find a root that isn't really a root... just remove it.

MrAdamBoyd (approved):
LGTM 🚢

PR #46, opened 2024-01-27
Correct cycle detection error ordering

dfed:
Instantiable cycles were being displayed backwards. Our test only had two elements in the cycle, so we weren't able to detect this.

### DIFF
--- a/Tests/SafeDIToolTests/SafeDIToolTests.swift
+++ b/Tests/SafeDIToolTests/SafeDIToolTests.swift
@@ -4182,7 +4182,7 @@ final class SafeDIToolTests: XCTestCase {
         await assertThrowsError(
             """
             Dependency cycle detected!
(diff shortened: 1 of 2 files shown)


Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -4182,40 +4182,45 @@ final class SafeDIToolTests: XCTestCase {
         await assertThrowsError(
             """
             Dependency cycle detected!
-            DefaultNetworkService -> DefaultLoggingService -> DefaultNetworkService
```
dfed, line 4185: this two-element cycle made it very difficult to determine whether the ordering of types in our error message was correct.

MrAdamBoyd (approved):
LGTM

PR #47, opened 2024-01-27
Ensure TypeDescription generates correct source for composite type when subtype has spaces

dfed:
After #42 we would improperly generate code for types that _must_ be wrapped within a single-element tuple. Example: an optional closure `() -> Void` would be generated as `() -> Void?` rather than `(() -> Void)?`. This PR ensures that we wrap types in a single-element tuple if they would otherwise be ambiguous.

While I was here, I also got us generating code that fulfills `ArrayTypeSyntax` and `DictionaryTypeSyntax` for the `.array` and `.dictionary` type descriptions.

### DIFF
--- a/Sources/SafeDICore/Models/TypeDescription.swift
+++ b/Sources/SafeDICore/Models/TypeDescription.swift
@@ -168,12 +168,13 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
     }
 
     public struct TupleElement: Codable, Hashable, Sendable {
+        init(label: String? = nil, _ typeDescription: TypeDescription) {
+            self.label = label
+            self.typeDescription = typeDescription
+        }
+
         public let label: String?
         public let typeDescription: TypeDescription
-
-        static func singleElement(_ typeDescription: TypeDescription) -> Self {
-            self.init(label: nil, typeDescription: typeDescription)
-        }
     }
 
     var isOptional: Bool {
@@ -253,7 +254,7 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
             self
         case .composition, .some, .any, .attributed, .closure:
             // These types contain spaces and may be ambigous without being wrapped.
-            .tuple([.singleElement(self)])
+            .tuple([.init(self)])
         }
     }
 }
@@ -333,7 +334,7 @@ extension TypeSyntax {
             let elements = typeIdentifier.elements.map {
                 TypeDescription.TupleElement(
                     label: $0.secondName?.text ?? $0.firstName?.text,
-                    typeDescription: $0.type.typeDescription
+                    $0.type.typeDescription
                 )
             }
             if elements.isEmpty {
@@ -436,7 +437,7 @@ extension ExprSyntax {
                 return .tuple(tupleElements.map {
                     TypeDescription.TupleElement(
                         label: $0.label?.text,
-                        typeDescription: $0.expression.typeDescription
+                        $0.expression.typeDescription
                     )
                 })
             }
--- a/Sources/SafeDICore/Models/TypeDescription.swift
+++ b/Sources/SafeDICore/Models/TypeDescription.swift
@@ -246,7 +246,7 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
         }
     }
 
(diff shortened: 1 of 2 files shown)


Tests/SafeDICoreTests/TypeDescriptionTests.swift
```diff
@@ -120,17 +120,28 @@ final class TypeDescriptionTests: XCTestCase {
 
     func test_typeDescription_whenCalledOnATypeSyntaxNodeRepresentingAOptionalTypeSyntax_findsTheType() throws {
         let content = """
-            protocol FooBar: Foo where Something == AnyObject? {}
+            var optionalAnyObject: AnyObject?
```
dfed, line 123: simplifying this test case allowed for a much simpler `OptionalTypeSyntaxVisitor` implementation

Tests/SafeDICoreTests/TypeDescriptionTests.swift
```diff
@@ -120,17 +120,28 @@ final class TypeDescriptionTests: XCTestCase {
...
-            guard optionalTypeIdentifier.asSource != "Something" else { continue }
-            XCTAssertFalse(optionalTypeIdentifier.isUnknown, "Type description is not of known type!")
-            XCTAssertTrue(optionalTypeIdentifier.asSource.contains("AnyObject?"))
-        }
+        let optionalTypeIdentifier = try XCTUnwrap(visitor.optionalTypeIdentifier)
+        XCTAssertFalse(optionalTypeIdentifier.isUnknown, "Type description is not of known type!")
+        XCTAssertEqual(optionalTypeIdentifier.asSource, "AnyObject?")
+    }
+
+    func test_typeDescription_whenCalledOnATypeSyntaxNodeRepresentingAOptionalClosureTypeSyntax_findsTheType() throws {
```
dfed, line 134: this test did not pass before the above changes!

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -240,6 +245,18 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
             return self
         }
     }
+
+    /// A representation of this type that is wrapped in a single element tuple to ensure cohesiveness of the type description. Can be used in source code.
```
dfed, line 249:
```suggestion
    /// A representation of this type that may be wrapped in a single element tuple to ensure cohesiveness of the type description.
```

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -240,6 +245,18 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
...
+    private var wrappedIfAmbiguous: Self {
+        switch self {
+        case .void, .simple, .optional, .implicitlyUnwrappedOptional, .metatype, .nested, .array, .dictionary, .tuple, .unknown:
+            // These types contain no spaces, and are therefore unambiguous without being wrapped.
+            self
+        case .composition, .some, .any, .attributed, .closure:
+            // These types contain spaces and may be ambigous without being wrapped.
+            .tuple([.init(self)])
+        }
+    }
```
MrAdamBoyd, line 259: I really like this. Simple.

MrAdamBoyd (approved):

Sources/SafeDICore/Models/TypeDescription.swift
dfed: took me a few tries to whittle it down to something this contained.

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -317,7 +334,7 @@ extension TypeSyntax {
             let elements = typeIdentifier.elements.map {
                 TypeDescription.TupleElement(
                     label: $0.secondName?.text ?? $0.firstName?.text,
-                    typeDescription: $0.type.typeDescription
+                    $0.type.typeDescription
```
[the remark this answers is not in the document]
dfed: let's me write `init(self)` which I liked 😅. There's also _always_ a type description in a tuple element so it seemed a little redundant to name it when it'll always be clear what it is.

I may waffle on this in a future PR, but for now I dig this.

PR #51, opened 2024-01-28
Enable instantiation cycles

### LINKED ISSUE #50: Enable instantiation cycles
**Goals**
* Enable the lazy instantiation of cycles of dependencies. i.e. it should be possible for A to lazily create B which lazily creates C which lazily creates A.

**Non-Goals**
* Enable the aggressive instantiation of cycles of dependencies. i.e. it should continue to not be possible for A to instantiate B which instantiates C which instantiates A.
* Enable the receiving of cycles of dependencies. i.e. it should continue to not be possible for A to receive B which receives C which receives A.

**Design**
Our generated code today has a fundamental flaw which does not enable the lazy instantiation of cycles of dependencies: properties that fulfill lazy instantiation can not be called from within their own definition. Here's how a type A which lazily creates B which lazily creates C which lazily creates A fails to compile with our current code generation:

```swift
extension Root {
    public convenience init() {
        let aBuilder = Instantiator<A> {
            let bBuilder = Instantiator<B> {
                let cBuilder = Instantiator<C> {
                    C(aBuilder: aBuilder) // Fails to compile! We can't reference aBuilder from within aBuilder's definition.
                }
                return B(cBuilder: cBuilder)
            }
            return A(bBuilder: bBuilder)
        }
        self.init(aBuilder: aBuilder)
    }
}
```

To enable the lazy creation of dependency cycles, we need a way to reference a builder from within the builder's definition. While a property can't be referenced from within a closure that instantiates the property, a function can. Here's how a type A which lazily creates B which lazily creates C which lazily creates A could be written in a way that compiles:

```swift
extension Root {
    public convenience init() {
        func __safeDI_aBuilder() -> A {
            func __safeDI_bBuilder() -> B {
                func __safeDI_cBuilder() -> C {
                    let aBuilder = Instantiator<A>(__safeDI_aBuilder)
                    return C(aBuilder: aBuilder)
                }
                let cBuilder = Instantiator<C>(__safeDI_cBuilder)
                return B(cBuilder: cBuilder)
            }
            let bBuilder = Instantiator<B>(__safeDI_bBuilder)
            return A(bBuilder: bBuilder)
        }
        let aBuilder = Instantiator<A>(__safeDI_aBuilder)
        self.init(aBuilder: aBuilder)
    }
}
```

When `ScopeGenerator`‘s `func generateCode` encounters an instantiator or forwarding instantiator property type, we should generate the builder method, and then declare the property that calls into the defined builder method. When a property declares a builder that forms the re-entry into the cycle, we should declare the property that calls into the builder method immediately, knowing that the call will compile + recurse.

Functions to fulfill forwarding instantiators will declare arguments matching the forwarded properties.

Where we detect and throw `ScopeError.dependencyCycleDetected` today, we'll instead need to create and return a `ScopeGenerator` that is aware that the builder function it needs to re-enter the cycle is already defined.

dfed:
Resolves #50. This PR is broken up into two commits for easier reviewing:

1. Makes lazy instantiated properties utilize functions rather than inline property declarations
2. Enables these functions to be called recursively to enable instantiation cycles 

Built on top of #46. Regardless of whether you review commit-by-commit, I **strongly recommend viewing the diff side-by-side, without showing whitespace changes**


Sources/SafeDICore/Models/Scope.swift
```diff
@@ -78,27 +78,31 @@ final class Scope: Hashable {
 
     func createScopeGenerator(
         for property: Property? = nil,
-        instantiableStack: OrderedSet<Instantiable> = [],
         propertyStack: OrderedSet<Property> = []
     ) throws -> ScopeGenerator {
-        if let cycleIndex = instantiableStack.firstIndex(of: instantiable) {
```
dfed, line 84: this was _way_ too aggressive a check before. We care about regenerating a property within a stack, not regenerating an instantiable (which may or may not be a constant, and may or may not have different property names)

Sources/SafeDICore/Models/Scope.swift
```diff
@@ -78,27 +78,31 @@ final class Scope: Hashable {
...
-            throw ScopeError.dependencyCycleDetected([instantiable] + instantiableStack.elements[0...cycleIndex])
+        if
+            let property,
+            property.propertyType.isConstant,
+            let cycleIndex = propertyStack.firstIndex(of: property)
+        {
+            throw ScopeError.dependencyCycleDetected([property.typeDescription] + propertyStack.elements[0...cycleIndex].map(\.typeDescription))
```
dfed, line 88: using property type descriptions instead of instantiables here isn't necessarily ideal since we no longer always reference the concrete instantiable type... but using the property type description makes it easier to find/follow the cycle in the code imo, so overall I think this is a win of a change

Sources/SafeDICore/Models/Scope.swift
```diff
@@ -78,27 +78,31 @@ final class Scope: Hashable {
...
                 childPropertyStack.insert(property, at: 0)
+                isPropertyCycle = propertyStack.contains(property)
+            } else {
+                isPropertyCycle = false
             }
             let scopeGenerator = ScopeGenerator(
                 instantiable: instantiable,
                 property: property,
-                propertiesToGenerate: try propertiesToGenerate.map {
+                propertiesToGenerate: isPropertyCycle ? [] : try propertiesToGenerate.map {
```
dfed, line 101: we don't need to generate properties if we're fulfilling a cycle – those properties are already fulfilled in the function that fulfills the cycle.

kierajmumick (approved):

dfed:
Force push is just a rebase with no code changes. Merging when green.

PR #59, opened 2024-02-04
Add support for type erasure with concrete existentials

### LINKED ISSUE #58: Support type erasure of SwiftUI View with AnyView
## Goals
* Enable dependency inversion when utilizing SwiftUI, like we do with UIKit
* Enable wrapping an `@Instantiated` property that is fulfilled by another type in the declared type-erased type's initializer when appropriate
* Enable wrapping a renamed `@Received` property in the declared type-erased type's initializer when appropriate
* Avoid adding more complexity than necessary

## Non-Goals
* Prevent consumers from writing code that will cause code-generation to fail to compile (we don't have enough type information to accomplish this)

## Investigations
Today, we support type erasure with UIKit by enabling a property to return a type further up in the inheritance tree than the object being created. e.g. it is possible to have a `Instantiator<UIViewController>` when we are actually creating a subclass of `UIViewController`

However, because SwiftUI both does not have an inheritance tree and has generics, a `Instantiator<View>` fails to compile. `Instantiator<any View>` can compile, but is useless since an `AnyView` can't take an `any View` in its initializer, and therefore `AnyView(myAnyViewInstantiator.instantiate())` fails to compile. Similarly, `Instantiator<some View>` fails to compile because there is no initialization clause.

It is standard practice in Swift libraries for an `Any*` type to have a no-label initializer. However, `AnyActor` is an exception to this standard practice.

### Initial design concept
The initial proposed design is simple: any `@Instantiated(fulfilledByType:)` or `@Received(fulfilledByDependencyNamed:ofType:)` property where the returned instance is a type that has an `Any` prefix will have the generated code that fulfills that type wrapped in the `AnyType`'s initializer.

In the following example, a type-erased `noteViewBuilder`
```swift
@Instantiated(fulfilledByType: "NoteView")
private let noteViewBuilder: ForwardingInstantiator<String, AnyView>
```

Would have the following code generated for it:
```swift
func __safeDI_noteViewBuilder(userName: String) -> AnyView {
    AnyView(NoteView(userName: userName, userService: userService, stringStorage: stringStorage))
}
```

We would know to do this because the concrete returned type of the forwarding instantiator is of type `AnyView`, which has the prefix `Any`. We also know that there is no `@Instantiatiable` type `AnyView` anywhere in the codebase.

While this design makes the strong assumption that any `Any`-prefixed type that is not `@Instantiable` is a type eraser with a no-label `public init(_:)` method, this is true of every Apple-made `Any`-prefixed type outside of `AnyActor`. Since making a strong assumption that is already proven incorrect is probably not the best approach, we move onto the next idea.

### Second design concept
If we want to avoid making a strong assumption, we can instead create two new macros:

```swift
public macro Instantiated(fulfilledByErasedType concreteTypeName: StaticString)
```

```swift
public macro Received<T>(fulfilledByDependencyNamed: StaticString, ofErasedType: T.Type)
```

When these new ErasedType macros are used, we know that we need to wrap the returned instance in our code generation in a no-label initializer of the erasing type.

In the following example, a type-erased `noteViewBuilder`
```swift
@Instantiated(fulfilledByErasedType: "NoteView")
private let noteViewBuilder: ForwardingInstantiator<String, AnyView>
```

Would have the following code generated for it:
```swift
func __safeDI_noteViewBuilder(userName: String) -> AnyView {
    AnyView(NoteView(userName: userName, userService: userService, stringStorage: stringStorage))
}
```

This design is more flexible than the initial design, but has the downside of adding yet another complexity layer to every autocomplete where it may not be necessary. I also don't love the `ErasedType` substring in the argument list.

### Third design concept
We can reduce the naming and autocomplete complexity in the above option by amending our existing macros to enable this new functionality:

```swift
public macro Instantiated(fulfilledByType concreteTypeName: StaticString, erasedToConcreteExistential: Bool = false)
```

```swift
public macro Received<T>(fulfilledByDependencyNamed: StaticString, ofType: T.Type, erasedToConcreteExistential: Bool = false)
```

When the new `erasedToConcreteExistential` macro parameters are used, we know that we need to wrap the returned instance in our code generation in a no-label initializer of the concrete erasing existential type.

In the following example, a type-erased `noteViewBuilder`
```swift
@Instantiated(fulfilledByType: "NoteView", erasedToConcreteExistential: true)
private let noteViewBuilder: ForwardingInstantiator<String, AnyView>
```

Would have the following code generated for it:
```swift
func __safeDI_noteViewBuilder(userName: String) -> AnyView {
    AnyView(NoteView(userName: userName, userService: userService, stringStorage: stringStorage))
}
```

This option has the benefits of both being explicit and having a sensible default option. Having a default option reduces complexity in the common case, which seems beneficial.

## Design

I am strongly leaning towards the third design concept.

## Note
When we make this change, we should update the README to use this rather than the non-compiling `some View`

dfed:
Resolves #58

This PR adds a parameter `erasedToConcreteExistential` to both the `@Instantiated` and `@Received` macros. This parameter enables the use of type-erasing concrete existentials – previously we only enabled erasing types to inherited types in both the `@Instantiated` and `@Received` macros. Since utilizing concrete existentials is a bit of a power feature, we have given this parameter a default argument of `false` – only folk who need the power of this argument will need to explore it.


README.md
```diff
@@ -177,8 +177,8 @@ public struct ParentView: View {
     // The Instantiator‘s `instantiate()` function will build a view of type `ChildView`.
     // All that is required for this code to compile is for there to be an
     // `@Instantiable public struct ChildView: View` in the codebase.
-    @Instantiated(fulfilledByType: "ChildView")
-    private let childViewBuilder: Instantiator<some View>
```
dfed, line 181: welp. this was never going to work

Sources/SafeDI/PropertyDecoration/Instantiated.swift
```diff
@@ -26,6 +26,26 @@
...
+///   - isExistential: Whether the decorated property’s type is an existential, or a concrete type that encapsulates a value conforming to a protocol without revealing the underlying type – type erasing types like `AnyView` fit this description. Set this parameter to `true` to encapsulate the created instance of `concreteTypeName` in a no-label initializer of the property’s type (e.g. `AnyView(ConcreteView())`).
 ///
-/// - Parameter concreteTypeName: The name of the concrete type that will be instantiated and assigned to this property. This parameter is only required when the decorated property’s type does not match an `@Instantiable` type or its `additionalTypes`. This parameter is particularly useful when working with a type-erased property.
-@attached(peer) public macro Instantiated(fulfilledByType concreteTypeName: StaticString = "") = #externalMacro(module: "SafeDIMacros", type: "InjectableMacro")
+///  - SeeAlso: [The Swift Programming Lanaguage’s explanation of existential types](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/opaquetypes/#Boxed-Protocol-Types)
```
dfed, line 47: this link is what inspired the name of the property `isExistential`

Sources/SafeDICore/Extensions/AttributeSyntaxExtensions.swift
```diff
@@ -76,13 +76,29 @@ extension AttributeSyntax {
         guard
             let arguments,
             let labeledExpressionList = LabeledExprListSyntax(arguments),
-            let lastLabeledExpression = labeledExpressionList.last,
```
dfed, line 79: we're no longer always the last expression!

Sources/SafeDICore/Models/Scope.swift
```diff
@@ -78,7 +78,8 @@ final class Scope: Hashable {
 
     func createScopeGenerator(
         for property: Property? = nil,
-        propertyStack: OrderedSet<Property> = []
+        propertyStack: OrderedSet<Property> = [],
+        erasedToConcreteExistential: Bool = false
```
dfed, line 82: we use a default value here (and in the other arguments) because root types call into this method, and root types are never erased.

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -169,22 +177,36 @@ actor ScopeGenerator: CustomStringConvertible {
...
+                        let instantiatedTypeDescription = property
+                            .typeDescription
+                            .unwrappedTypeDescription
+                            .asInstantiatedType
+                            .asSource
                         let propertyDeclaration = if typeDescription == unwrappedTypeDescription {
                             "let \(property.label)"
                         } else {
-                            "let \(property.label): \(typeDescription)"
+                            "let \(property.asSource)"
```
dfed, line 200: not a functional change, but this change makes our code generation a little more consistent / less repetitive.

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -501,7 +501,7 @@ final class SafeDIToolTests: XCTestCase {
         )
     }
 
-    func test_run_writesConvenienceExtensionOnRootOfTree_whenRootInstantiatesPropertiesThatUtilizedSingleForwardedPropertyInSubBuilders() async throws {
```
dfed, line 504: while I was here I realized that we were using the phrase `Utilized` when we meant `Utilizes` in our test names

MrAdamBoyd (approved):
I had no idea what this was called before reading this.

PR #62, opened 2024-02-21
Enable code-gen to compile when imports have attributes

dfed:
As title

### DIFF
--- a/Sources/SafeDICore/Extensions/ImportDeclSyntaxExtensions.swift
+++ b/Sources/SafeDICore/Extensions/ImportDeclSyntaxExtensions.swift
@@ -54,8 +54,8 @@ extension ImportDeclSyntax {
 
     private var attribute: String {
         // This AttributeList is of the form ["@", "attribute"]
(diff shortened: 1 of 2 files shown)


Sources/SafeDICore/Extensions/ImportDeclSyntaxExtensions.swift
```diff
@@ -54,8 +54,8 @@ extension ImportDeclSyntax {
 
     private var attribute: String {
         // This AttributeList is of the form ["@", "attribute"]
-        // So we grab the last token
-        attributes.lastToken(viewMode: .sourceAccurate)?.text ?? ""
```
dfed, line 58: I'd stolen this bit of code [from SwiftInspector](https://github.com/fdiaz/SwiftInspector/blob/450c62a742b318f9b45643de37e8ef29a6c840f1/Sources/SwiftInspectorVisitors/ImportVisitor.swift#L123-L135) and didn't read what it was doing carefully enough 😬

PR #66, opened 2024-04-06
Remove ForwardingInstantiator in favor of ErasedInstantiator

### LINKED ISSUE #44: ForwardingInstantiator typealias for easier declarations
**Goals**
Take an example class:
```
@Instantiable
public final class MetricsService {
  @Forwarded
  private let user
}
```

If a class wants to instantiate `MetricsService`, they must use this declaration:
```
@Instantiated
private let metricsServiceBuilder: ForwardingInstantiator<MetricsService.ForwardedArguments, MetricsService>
```

However, SafeDI can easily generate another typealias, similar to `forwardedArguments` which makes this declaration shorter:
```
@Instantiated
private let metricsServiceBuilder: MetricsService.ForwardingInstantiator
```

Which would look like this in the codegen:
```
extension MetricsService {
  typealias ForwardingInstantiator = FowardingInstantiator</* forwarded types here */, MetricsService> 
}
```

dfed:
This PR resolves #44 by making it such that you no longer need to specify forwarded properties on an instantiator unless you are utilizing type erasure. This approach makes simple work simple, and reserves complexity for the complex tasks.

In order to accomplish this, I made the following changes to the API:

1. `Instantiator` now takes a type that conforms to `Instantiable`, where `Instantiable` is a type that has an `associatedtype ForwardedProperties`.
2. The `@Instantiable` macro requires conformance to the `Instantiable` protocol via a FixIt (we can't generate it in all cases)

This is a breaking change.


Sources/SafeDICore/Errors/FixableInstantiableError.swift
```diff
@@ -21,19 +21,16 @@
...
     case dependencyHasTooManyAttributes
     case dependencyHasInitializer
     case missingPublicOrOpenAttribute
     case missingRequiredInitializer(hasInjectableProperties: Bool)
-    case missingRequiredInstantiateMethod(typeName: String)
-    case missingAttributes
-    case disallowedGenericParameter
-    case disallowedEffectSpecifiers
-    case incorrectReturnType
-    case disallowedGenericWhereClause
```
dfed, line 33: These moved to the `FixableInstantiableExtensionError` type

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -39,7 +39,7 @@ public final class DependencyTreeGenerator {
 
         let typeDescriptionToScopeMap = try createTypeDescriptionToScopeMapping()
         try validateReceivedProperties(typeDescriptionToScopeMap: typeDescriptionToScopeMap)
-        let rootScopeGenerators = try rootInstantiableTypes
+        let rootScopeGenerators = try rootInstantiables
```
dfed, line 42: find/replace casualty but I don't mind it.

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -224,10 +224,10 @@ public final class InstantiableVisitor: SyntaxVisitor {
                     )
                 )
             )
-            modifiedNode.funcKeyword.leadingTrivia = .spaces(0)
+            modifiedNode.funcKeyword.leadingTrivia = []
```
dfed, line 227: the proper way to set no trivia. TIL

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
MrAdamBoyd: I don't either. I also like now that you can return `Instantiable` instead of `InstaniableType` elsewhere.

Sources/SafeDIMacros/Macros/InstantiableMacro.swift
```diff
@@ -30,7 +30,11 @@ public struct InstantiableMacro: MemberMacro {
         providingMembersOf declaration: some DeclGroupSyntax,
         in context: some MacroExpansionContext
     ) throws -> [DeclSyntax] {
-        if let fulfillingAdditionalTypesArgument = declaration.attributes.instantiableMacro?.fulfillingAdditionalTypes {
+        if
+            let fulfillingAdditionalTypesArgument = (
+                declaration.attributes.instantiableMacro
+            )?.fulfillingAdditionalTypes
```
MrAdamBoyd, line 36: Did you mean to have the `let` on a new line?

MrAdamBoyd (approved):
LGTM

Sources/SafeDIMacros/Macros/InstantiableMacro.swift
dfed: Yeah when the line wraps like this I like to give it its own visual space

Issue #65, opened 2024-03-07
nonisloated init produces warning on classes decorated with @MainActor


MrAdamBoyd:
**Steps to reproduce**
1. Upgrade to Xcode 15.3
2. Try to build

**Sample code or sample project upload**
```
@MainActor
@Instantiable
public final class ApplicationCoordinator: NSObject {
    private(set) var window: UIWindow?

    // MARK: - Initializers
    nonisolated
    public init(
        featureFlagManager: FeatureFlagManager
    ) {
        self.featureFlagManager = featureFlagManager
        super.init()
        // Do other work here
    }

    @Instantiated
    private let featureFlagManager: FeatureFlagManager
```

Now results in an error:
`Main actor-isolated property featureFlagManager can not be mutated from a non-isolated context; this is an error in Swift 6`

dfed:
I think we may need to remove `nonisolated` from our initializers, and add `@MainActor` to the root init if our root is also `@MainActor`. I need to play with this, but it should be doable. Also happy to have help on this one. I'm a few weeks from upgrading to Xcode 15.3 since I have other dependencies that are problematic.

dfed:
Also need to keep track of which builders build main-actor-bound types and mark those closures (and `instantiate` methods) as main-actor too

dfed:
Once #66 lands, I'll be able to tackle this by creating:
```
public final class MainInstantiator<T: Instantiable> { ... }
public final class MainErasedInstantiator<T: Instantiable> { ... }
```

We'll require using a `Main*Instantiator` if any (recursive) child instantiated property is `@MainActor`-bound.

I need to think a bit about how we will try to detect that a type is `@MainActor`-bound, since it is possible for `@MainActor` to be inherited from a superclass or protocol conformance. I am currently leaning towards not trying to detect this, and instead writing good documentation and relying on build errors to guide folk. Once you've made a `Main*Instantiator`, though, we'll know that anything that builds that must also be a `Main*Instantiator`.

dfed:
New naming proposal:

```
public final class Instantiator<T: Instantiable> { ... }
public final class ErasedInstantiator<T: Instantiable> { ... }

public final class UnboundInstantiator<T: Instantiable> { ... }
public final class UnboundErasedInstantiator<T: Instantiable> { ... }
```

In this proposal, we're main-by-default. If you want to be able to instantiate types off of main, you use an `Unbound*Instantiator`.

PR #70, opened 2024-04-07
Enable instantiating @MainActor-bound types in Swift 5.10+

dfed:
This PR makes it possible to use SafeDI with Swift 5.10+ and Strict Concurrency checking turned on. To accomplish this, we made the following changes:

1. Made generated `init` methods on `@Instantiable` be isolated to the parent actor context – removed the `nonisolated` from these generated `init` methods.
1. The `instantiate(...)` method for both `Instantiator` and `ErasedInstantiator` become `@MainActor`-bound
1. Created two new instantiator objects: `NonisolatedInstantiator` and `NonisolatedErasedInstantiator`. As the name implies, these instantiators's `instantiate(...)` methods run in a nonisolated context.

We made `Instantiator` and `ErasedInstantiator` `@MainActor`-bound because in most applications the root of the dependency tree is `@MainActor`-bound, and in applications the majority of delayed-built products are `@MainActor`-bound. In an effort to make the common case simple, we have optimized for simplifying the `@MainActor`-bound case.

Addresses #65.

This is a breaking change.

To adopt this change locally, consumers will need to follow their build errors after updating to 0.5.0. If something requires instantiation outside of main actor isolation, utilize a `Nonisolated`-prefixed Instantiator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ public struct MyApp: App, Instantiable {
 }
 ```
 
-An `Instantiator` is `@MainActor`-bound – if you want to instantiate a type in a nonisolated environment, use an [`NonisolatedInstantiator`]](Sources/SafeDI/DelayedInstantiation/NonisolatedInstantiator.swift).
+An `Instantiator` is `@MainActor`-bound – if you want to instantiate a type in a nonisolated environment, use a [`NonisolatedInstantiator`]](Sources/SafeDI/DelayedInstantiation/NonisolatedInstantiator.swift).
 
 #### Utilizing @Instantiated with type erased properties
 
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -156,36 +156,6 @@ Property declarations within `@Instantiable` types decorated with [`@Instantiate
 
 `@Instantiated`-decorated properties must be an `@Instantiable` type, or of an `additionalType` listed in an `@Instantiable(fulfillingAdditionalTypes:)`‘s declaration.
 
-#### Utilizing @Instantiated with type erased properties
```
dfed, line 159: this moved down so that we discuss it _after_ we discuss the `Instantiator`

Sources/SafeDI/DelayedInstantiation/ErasedInstantiator.swift
```diff
@@ -19,34 +19,40 @@
 // SOFTWARE.
 
 /// A SafeDI dependency designed for the deferred instantiation of a type-erased instance of a
-/// type decorated with `@Instantiable`. Instantiation is thread-safe.
```
dfed, line 22: Given that we're explicitly determining threading here it seemed duplicative to say `Instantiation is thread-safe.`

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -1974,7 +1974,7 @@ final class SafeDIToolTests: XCTestCase {
 
             extension Root {
                 public convenience init() {
-                    func __safeDI_childABuilder(recreated: Recreated) -> ChildA {
+                    nonisolated func __safeDI_childABuilder(recreated: Recreated) -> ChildA {
```
dfed, line 1977: this part of the code-gen has me pretty convinced that `Nonisolated` was indeed the correct prefix on the `Instantiator`

Sources/SafeDI/DelayedInstantiation/NonisolatedErasedInstantiator.swift
```diff
@@ -0,0 +1,56 @@
+// Distributed under the MIT License
```
MrAdamBoyd, line 1: How I wish you could `git cp`

Sources/SafeDICore/Models/Property.swift
```diff
@@ -136,33 +136,48 @@ public struct Property: Codable, Hashable, Comparable, Sendable {
         /// An `Instantiator` property.
         /// The instantiated product is not forwarded down the dependency tree. This is done intentionally to avoid unexpected retains.
         case instantiator
-        /// A `ErasedInstantiator` property.
+        /// An `ErasedInstantiator` property.
```
MrAdamBoyd, line 139: Copy paste error? 😛

Tests/SafeDIMacrosTests/InjectableMacroTests.swift
```diff
@@ -169,7 +169,7 @@ final class InjectableMacroTests: XCTestCase {
             public struct ExampleService {
                 @Instantiated
                 ┬────────────
-                ╰─ 🛑 An `ErasedInstantiator` cannot be used without the argument `fulfilledByType`
+                ╰─ 🛑 `ErasedInstantiator` and `NonisolatedErasedInstantiator` cannot be used without the argument `fulfilledByType`
```
MrAdamBoyd, line 172: Double negative

```
`ErasedInstantiator` and `NonisolatedErasedInstantiator` require the argument `fulfilledByType`
```
?

MrAdamBoyd (approved):
LGTM. Just one grammar question.

Sources/SafeDICore/Models/Property.swift
dfed: most definitely

Sources/SafeDI/DelayedInstantiation/NonisolatedErasedInstantiator.swift
dfed: if only!

PR #71, opened 2024-04-09
Fix a minor Markdown syntax error in README

bachand:
I noticed that this link was not rendering properly when reading the README.


dfed (approved):
thank you!!

PR #77, opened 2024-04-22
Explicitly disallow optionals in 'fulfillingAdditionalTypes' parameter

dfed:
We've never supported a type fulfilling an optional type and not the underlying type. Here we make that explicit.


Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -777,14 +777,46 @@ final class InstantiableMacroTests: XCTestCase {
         assertMacro {
             """
             @Instantiable
-            public enum ExampleService {}
+            public enum ExampleService: Instantiable {}
```
dfed, line 780: added these `Instantiable` conformances while I was here. It helps make this test more resilient to underlying code ordering.

PR #82, opened 2024-06-10
Support @retroactive conformance in Swift 6

dfed:
Swift 6 introduces a new warning that you must `@retroactive`-ly apply conformance in an extension on a type you do not own. This breaks SafeDI's code parsing. This PR gets SafeDI working great in Swift 6.


Sources/SafeDIMacros/Macros/InstantiableMacro.swift
```diff
@@ -365,3 +357,28 @@ public struct InstantiableMacro: MemberMacro {
...
+            )
+            || self == .nested(
+                name: "Instantiable",
+                parentType: .attributed(
+                    .simple(name: "SafeDI"),
+                    specifier: nil,
+                    attributes: ["retroactive"]
+                )
+            )
+    }
```
dfed, line 379: this clearly does not scale but it works well enough for now.

MrAdamBoyd: Seems reasonable safe for now. If another change is needed for Swift 7 then something else could be done.

MrAdamBoyd (approved):

PR #88, opened 2024-06-23
Enable generic types to be instantiated

### LINKED ISSUE #86: Enable generic types to be instantiated
Today, a `Container<Element>` type cannot effectively be `@Instantiable` or `@Instantiated` unless an `init` method can be written that does not utilize the generic `Element`

Ideally, it would be possible for a type with a generic to be `@Instantiable`. Imagine something like:

```swift
@Instantiable
extension Container: Instantiable {
    public static func instantiate() -> Container<Bool> {
        .init(contained: false)
    }

    public static func instantiate() -> Container<Int> {
        .init(contained: 0)
    }

    public static func instantiate() -> Container<String> {
        .init(contained: "")
    }
}
```

Accomplishing this should be relatively straight-forward. We currently have a `FixableInstantiableError.incorrectReturnType` that throws whenever the return type's description is not the same as the extended type. We could make this error more permissive and not throw as long as the base name (i.e. the type name without generics) is the same.

I don't see a good way to make generics instantiable outside of an `extension`, but I am open to ideas!

dfed:
Resolves #86


Sources/SafeDI/PropertyDecoration/Instantiable.swift
```diff
@@ -40,7 +40,7 @@
 ///     }
 ///
 /// An extension declaration decorated with `@Instantiable` makes the extended type capable of having properties of other `@Instantiable`-decorated types injected into it. Decorating extensions with `@Instantiable` enables third-party types to be instantiated by the SafeDI system.
-/// Usage of this macro requires the extension to implement a method `public static instantiate() -> ExtendedType` that defines the instantiation logic for the externally defined type.
+/// Usage of this macro requires the extension to implement a method `public static func instantiate() -> ExtendedType` that defines the instantiation logic for the externally defined type.
```
dfed, line 43: 🤦‍♂️

Tests/SafeDIToolTests/SafeDIToolTests.swift
```diff
@@ -4659,7 +4715,7 @@ final class SafeDIToolTests: XCTestCase {
 
                     @Instantiable
                     extension RootViewController: Instantiable {
-                        public static instantiate() {
```
dfed, line 4662: 😅 kinda impressed these tests were working despite the obvious syntax errors in the test code

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -138,16 +138,10 @@ public final class InstantiableVisitor: SyntaxVisitor {
     }
 
     public override func visit(_ node: ExtensionDeclSyntax) -> SyntaxVisitorContinueKind {
-        guard declarationType.isExtension else {
-            return .skipChildren
-        }
```
dfed, line 143: really not necessary. Extensions can't be nested in other types, and we only call the `InstantiableVisitor` with an `ExtensionDeclSyntax` on extensions.

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -138,16 +138,10 @@ public final class InstantiableVisitor: SyntaxVisitor {
...
-        guard let instantiableMacro = node.attributes.instantiableMacro else {
-            // Not an instantiable type. We do not care.
-            return .skipChildren
-        }
```
dfed, line 147: again, this was super dead code – no way to reach it given how it's called. we were protecting against nothing.

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -317,29 +317,24 @@ public final class InstantiableVisitor: SyntaxVisitor {
         }
     }
 
-    // MARK: Internal
-
-    var instantiable: Instantiable? {
-        guard let instantiableType else { return nil }
+    public var instantiables: [Instantiable] {
```
dfed, line 320: now extensions can create _multiple_ `Instantiable` types, since `Container<Int>` and `Container<String>` are considered distinct Instantiable types by SafeDI, even though they are still a single generic type to the Swift compiler.

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -244,6 +244,17 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
         }
     }
 
+    var strippingGenerics: TypeDescription {
+        switch self {
+        case let .simple(name, _):
+            .simple(name: name, generics: [])
+        case let .nested(name, parentType, _):
+            .nested(name: name, parentType: parentType, generics: [])
+        case .array, .attributed, .any, .closure, .composition, .dictionary, .implicitlyUnwrappedOptional, .metatype, .optional, .some, .tuple, .unknown, .void:
```
dfed, line 253: it's possible to strip generics out of some of these types (like optionals)... but doing so doesn't really mean anything to SafeDI today. Keeping this on the simple side until we need it.

NickEntin (approved):

PR #89, opened 2024-06-23
Do not throw errors when declaring multiple instantiate() methods with multiple generic types in a single extension

dfed:
Fills in a gap I missed in #88


Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -211,7 +210,7 @@ actor ScopeGenerator: CustomStringConvertible {
                             .unwrappedTypeDescription
                             .asInstantiatedType
                             .asSource
-                        let propertyDeclaration = if typeDescription == unwrappedTypeDescription {
+                        let propertyDeclaration = if !instantiable.declarationType.isExtension, typeDescription == unwrappedTypeDescription {
```
dfed, line 213: this check and the check below are necessary to ensure our generated code is not ambiguous

Sources/SafeDIMacros/Macros/InstantiableMacro.swift
```diff
@@ -209,10 +209,17 @@ public struct InstantiableMacro: MemberMacro {
                 context.diagnose(diagnostic)
             }
 
-            let instantiableTypeCount = visitor.instantiables.map(\.instantiableTypes).count
-            if instantiableTypeCount > 1 {
```
dfed, line 213: oof. this was the wrong check previously. and our error was not the most legible

PR #95, opened 2024-07-01
Improve error messages when receiving a property that forms a cycle

dfed:
Resolves #93 and #94

PR #100, opened 2024-07-07
Improve + consolidate cycle detection

dfed:
Builds on top of #95 by fully consolidating our cycle detection + further improving the error message we emit when receiving a property that isn't yet fully instantiated.

### DIFF
--- a/Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
+++ b/Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
@@ -353,13 +353,12 @@ public final class DependencyTreeGenerator {
             if let property {
                 func validateNoCycleInReceivedProperties(
                     scope: Scope,
-                    instantiatedProperty: Property,
                     receivedPropertyStack: OrderedSet<Property>
                 ) throws {
                     for childProperty in scope.receivedProperties {
-                        guard childProperty != instantiatedProperty else {
+                        guard childProperty != property else {
                             throw DependencyTreeGeneratorError.receivedConstantCycleDetected(
-                                instantiated: instantiatedProperty,
+                                instantiated: property,
                                 receivedPropertyChain: receivedPropertyStack + [childProperty]
                             )
                         }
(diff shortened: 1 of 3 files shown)


Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -389,16 +433,17 @@ public final class DependencyTreeGenerator {
...
                         continue
                     }
-                    try validateReceivedProperties(
+                    try validatePropertiesAreFulfillable(
                         on: childScope,
                         receivableProperties: receivableProperties
-                            .union(scope.properties),
-                        property: property,
+                            .union(scope.properties)
+                            .removing(childProperty),
```
dfed, line 445: brought this line back from #95. Now that we're doing our cycle checking before we're doing our "initialized in chain" checking, we can make this more semantically correct

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -351,58 +351,43 @@ actor ScopeGenerator: CustomStringConvertible, Sendable {
     private var generateCodeTask: Task<String, Error>?
 
     private var orderedPropertiesToGenerate: [ScopeGenerator] {
-        get throws {
```
dfed, line 354: we no longer need to `throw` here since we do this checking much earlier now. best reviewed with whitespace turned off

Tests/SafeDIToolTests/SafeDIToolCodeGenerationErrorTests.swift
```diff
@@ -960,8 +960,59 @@ final class SafeDIToolCodeGenerationErrorTests: XCTestCase {
     func test_run_onCodeWithCircularReceivedRenamedDependencies_throwsError() async {
         await assertThrowsError(
             """
-            Dependency cycle detected on A!
-            b: B -> c: C -> renamedB: B -> b: B
+            Dependency received in same chain it is instantiated!
+            @Instantiated a: A -> @Received renamedB: B -> @Received c: C -> @Received a: A
```
dfed, line 964: this test better mirrors the above test. I liked the parallelism

PR #102, opened 2024-07-07
Improve test coverage calculation

dfed:
So our test coverage has been cheating a bit – we were including our `Tests/` code in our coverage numbers. Now we aren't. To get us back to 99.5+% coverage I added a few more tests. I also mucked with the shape of some lines to make it easier to get coverage – I'm not proud of it... but... here we are.


Sources/SafeDICore/Extensions/ImportDeclSyntaxExtensions.swift
```diff
@@ -23,14 +23,11 @@ import SwiftSyntax
 extension ImportDeclSyntax {
     // MARK: Public
 
-    public var asImportStatement: ImportStatement? {
-        guard let moduleName = path.first?.name.text else {
```
dfed, line 27: we really don't need to `guard` here – if our code-gen is running then our build succeeded, which means we don't have any dangling `import` statements.

Sources/SafeDICore/Errors/FixableInjectableError.swift
```diff
@@ -41,38 +41,32 @@ public enum FixableInjectableError: DiagnosticError {
     // MARK: - InjectableDiagnosticMessage
 
     private struct InjectableDiagnosticMessage: DiagnosticMessage {
-        var diagnosticID: MessageID {
```
dfed, line 44: these computed properties were never executed in tests. There's no good way to do so as part of an integration test. So now we're just creating them up-front.

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -105,7 +105,9 @@ public final class DependencyTreeGenerator {
                         .map(\.asSource)
                         .joined(separator: " -> "))
                     """
-                }.joined(separator: "\n"))
+                }
+                .sorted()
```
dfed, line 109: we didn't have a test for multiple fulfillable properties! once I added one, it became clear that the error wasn't stable, so we now sort here.

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -136,11 +138,7 @@ public final class DependencyTreeGenerator {
             }
         }
 
-        struct UnfulfillableProperty: Hashable, Comparable {
```
dfed, line 139: we don't need the property to be comparable because the generated error is already comparable and we sort that

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -153,32 +151,25 @@ public final class DependencyTreeGenerator {
     private let typeDescriptionToFulfillingInstantiableMap: [TypeDescription: Instantiable]
     private var rootScopeGenerators: [ScopeGenerator] {
         get throws {
-            if let _rootScopeGenerators {
```
dfed, line 156: we never actually used the cached object. so now it's gone.

Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -265,12 +256,9 @@ public final class DependencyTreeGenerator {
         // Create the mapping.
         let typeDescriptionToScopeMap: [TypeDescription: Scope] = reachableTypeDescriptions
             .reduce(into: [TypeDescription: Scope]()) { partialResult, typeDescription in
-                guard let instantiable = typeDescriptionToFulfillingInstantiableMap[typeDescription] else {
-                    // We can't find an instantiable for this type.
-                    // This is bad, but we handle this error in `validateReachableTypeDescriptions()`.
-                    return
```
dfed, line 271: this was impossible to hit, and we already have `validateReachableTypeDescriptions()` to handle less egregious cases. So I folded in this `guard let` with the one below

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -222,28 +222,6 @@ public enum TypeDescription: Codable, Hashable, Comparable, Sendable {
         }
     }
 
-    var isUnknown: Bool {
```
dfed, line 225: this was only used in tests – so I moved it to tests

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -423,8 +401,6 @@ extension ExprSyntax {
             if memberAccessExpr.declName.baseName.text == "self" {
                 if let base = memberAccessExpr.base {
                     return base.typeDescription
-                } else {
-                    return .unknown(text: memberAccessExpr.trimmedDescription)
```
dfed, line 427: unreachable

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -440,8 +416,6 @@ extension ExprSyntax {
                             generics: []
                         )
                     }
-                } else {
-                    return .unknown(text: memberAccessExpr.trimmedDescription)
```
dfed, line 444: unreachable

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -508,8 +482,6 @@ extension ExprSyntax {
                     doesThrow: arrow.effectSpecifiers?.throwsSpecifier != nil,
                     returnType: returnType.typeDescription
                 )
-            } else {
-                return .unknown(text: trimmedDescription)
```
dfed, line 512: unreachable

Sources/SafeDICore/Models/TypeDescription.swift
```diff
@@ -529,9 +501,8 @@ extension ExprSyntax {
                 key: onlyElement.key.typeDescription,
                 value: onlyElement.value.typeDescription
             )
-        } else {
-            return .unknown(text: trimmedDescription)
         }
+        return .unknown(text: trimmedDescription)
```
dfed, line 505: I hate fall through like this but... it does work

Tests/SafeDIToolTests/SafeDIToolCodeGenerationTests.swift
```diff
@@ -2685,7 +2685,7 @@ final class SafeDIToolCodeGenerationTests: XCTestCase {
                 }
                 """,
                 """
-                import Foundation
+                @_implementationOnly import Foundation
```
dfed, line 2688: helps exercise some of our import handling code we previously were not

PR #104, opened 2024-07-22
Generate required initializer via FixIt instead of code-generation

dfed:
As I continue to utilize SafeDI in a production application, I keep hitting a wall: macro-generated code can not be referenced by other macros. This is particularly difficult when utilizing the `#Preview` macro with an `@Instantiable` type: the generated initializer **can not** be utilized by `#Preview`, which means that the `@Instantiable`-generated `init` is often unusable and therefore useless.

This PR gets us away from the `@Instantiable` macro generating initializers. Instead: `@Instantiable` now generates an error with a FixIt to create the `init` method that previously would have been generated.

I don't love this change. There are few concrete downsides:
* FixIts are less helpful than macro-generated code, since macro-generated code can auto-update the existing generated `init` when dependencies change. FixIts can help you generate the correct code, but don't delete existing incorrect code.
* FixIts require you to pause compilation and fix things, whereas generated code does not.

The upside, of course, is that these fixit-generated initializers are usable **everywhere**, including in `#Preview` and other macros.

**This is a breaking change**


Sources/SafeDICore/Models/Property.swift
```diff
@@ -83,7 +83,10 @@ public struct Property: Codable, Hashable, Comparable, Sendable {
                         ))
                         if let attributes {
                             for attribute in attributes {
-                                AttributeSyntax(attributeName: IdentifierTypeSyntax(name: .identifier(attribute)))
+                                AttributeSyntax(
+                                    attributeName: IdentifierTypeSyntax(name: .identifier(attribute)),
+                                    trailingTrivia: .space
```
dfed, line 88: this has been broken for a bit I think! generated code that compiles but it looked bad.

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -33,14 +33,14 @@ public final class InstantiableVisitor: SyntaxVisitor {
...
-            declarationType.isTypeDefinition
-            && node.modifiers.staticModifier == nil
+            declarationType.isTypeDefinition,
+            node.modifiers.staticModifier == nil
         else {
             return .skipChildren
         }
         // Check attributes and extract dependency source.
         let dependencySources = node.attributes.dependencySources
-        guard dependencySources.isEmpty || dependencySources.count == 1 else {
```
dfed, line 43: this was just plain wrong 😅

Sources/SafeDICore/Visitors/InstantiableVisitor.swift
```diff
@@ -33,14 +33,14 @@ public final class InstantiableVisitor: SyntaxVisitor {
...
+        if dependencySources.count > 1 {
```
dfed, line 43: we no longer `guard` because we want to fall through now and generate all the diagnostics below where applicable.

Sources/SafeDIMacros/Macros/InstantiableMacro.swift
```diff
@@ -106,20 +106,29 @@ public struct InstantiableMacro: MemberMacro {
                 .contains(where: { $0.isValid(forFulfilling: visitor.dependencies) })
             guard hasMemberwiseInitializerForInjectableProperties else {
                 if visitor.uninitializedNonOptionalPropertyNames.isEmpty {
-                    var initializer = Initializer.generateRequiredInitializer(
```
dfed, line 109: I recommend viewing this file in side-by-side mode – this code mostly moved around rather than changed

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -44,78 +44,272 @@ import SafeDICore
             }
         }
 
-        // MARK: Generation tests
```
dfed, line 47: we no longer generate code. a ton of tests moved down.

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -44,78 +44,272 @@ import SafeDICore
...
+        // MARK: Error tests
+
+        func test_declaration_throwsErrorWhenOnProtocol() {
```
dfed, line 49: error tests moved up without changes

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -44,78 +44,272 @@ import SafeDICore
...
+                """
+                @Instantiable(fulfillingAdditionalTypes: [AnyObject?.self])
+                ┬──────────────────────────────────────────────────────────
+                ╰─ 🛑 The argument `fulfillingAdditionalTypes` must not include optionals
+                public final class ExampleService: Instantiable {}
+                """
+            }
+        }
 
-        func test_declaration_generatesRequiredInitializerWithoutAnyDependenciesOnStruct() {
```
dfed, line 49: copy/paste – this name was just wrong

dfed:
Ah shucks. I need to do a documentation pass before merging. I'm pretty sure I mentioned generating initializers in the README. Should be a simple change.

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -387,50 +642,116 @@ import SafeDICore
...
+                    let instantiatedA: InstantiatedA
+                }
+                """
+            } fixes: {
+                """
+                @Instantiable
+                public struct ExampleService: Instantiable {
+                public init(instantiatedA: InstantiatedA) {
+                self.instantiatedA = instantiatedA
+                }
```
NickEntin, line 682: Expected that indentation on a lot of these fixits looks off?

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -626,40 +1023,79 @@ import SafeDICore
...
+                self.forwardedA = forwardedA
+                self.receivedA = receivedA
+                self.receivedB = receivedB
+                }
+
                     public init(forwardedA: ForwardedA, receivedA: ReceivedA) {
                         self.forwardedA = forwardedA
                         self.receivedA = receivedA
                         receivedB = ReceivedB()
                     }
```
NickEntin, line 1060: Could this case have the fixit add to the existing init instead of creating a new one?

NickEntin (approved):

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
dfed: yeah – Xcode corrects it. I explicitly don't want to manage indentation in fixits because I don't know the end-user's preferences.

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
dfed: technically yes, but:
1. There are times that this updating an existing `init` won't be what is actually desired.
2. It's a good bit harder to find the closest `init` and suggest updating it. Doable, but harder.
3. Deleting code is easier than writing it.

Sources/SafeDIMacros/Macros/InstantiableMacro.swift
```diff
@@ -106,20 +106,29 @@ public struct InstantiableMacro: MemberMacro {
...
-                    return [DeclSyntax(initializer)]
-                        + generateForwardedProperties(from: forwardedProperties)
+                    context.diagnose(Diagnostic(
+                        node: Syntax(declaration.memberBlock),
+                        error: FixableInstantiableError.missingRequiredInitializer(.hasOnlyInjectableProperties),
+                        changes: [
+                            .replace(
+                                oldNode: Syntax(declaration.memberBlock.members),
+                                newNode: Syntax(membersWithInitializer)
+                            ),
```
kierajmumick, line 128: @dfed If there is an initializer that fits the pattern of the generated init, you could replace _only_ that initializer with a new initializer rather than adding an initializer into the member block. 

This could alleviate some of the pain of having the initializer not auto-generated with macros.

dfed: Yup. The issue with doing this is that I only want to emit one and only one error per problem, and I can't be sure which initializer to update if there are many. I can't add multiple fixits to a single error.

Maybe I'm overthinking it and having multiple errors would be okay? But it feels wrong to me.

PR #107, opened 2024-08-22
Xcode 16.0b6 fixes

dfed:
Gets us compiling without warnings on the latest Xcode 16.0 beta!

### DIFF
--- a/Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
+++ b/Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
@@ -20,7 +20,7 @@
 
 import Collections
 
-public final class DependencyTreeGenerator {
+public actor DependencyTreeGenerator {
     // MARK: Initialization
 
     public init(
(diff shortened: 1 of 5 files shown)


Sources/SafeDICore/Generators/DependencyTreeGenerator.swift
```diff
@@ -20,7 +20,7 @@
 
 import Collections
 
-public final class DependencyTreeGenerator {
+public actor DependencyTreeGenerator {
```
dfed, line 23: We were getting an error that access to `DependencyTreeGenerator` was racy. Turns out all the API here is already async, so might as well make this type an `actor`

Sources/SafeDITool/SafeDITool.swift
```diff
@@ -23,7 +23,7 @@ import Foundation
 import SafeDICore
 import SwiftParser
 #if canImport(ZippyJSON)
-    @preconcurrency import ZippyJSON
```
dfed, line 26: honestly feels like a bug that `ZippyJSONDecoder` isn't treated as Sendable given that I was doing a `@preconcurrency import` but... oh well.

PR #108, opened 2024-09-01
Suggest likely alternative receivable properties when received property is unfulfillable as written

### LINKED ISSUE #106: Provide a better error when type names are not equally-qualified in instantiated and received declarations
**Goals**
Make it less painful to debug when you refer to a type in an `Instantiator` by two different names. 

**Design**
Give a more actionable error when we have a type that doesn't match anything, but there's an existing type that is the same name prefixed by another type. For example:

```swift
@Instantiated private let someThing: Instantiated<View>

@Received private let someThing: Instantiated<MyViewController.View>
```

dfed:
Resolves #106


NickEntin (approved):

NickEntin (commented):
Maybe add a test for optional/non-optional suggestion

PR #109, opened 2024-09-02
Enable Xcode 16 plugin to use versioned release build

dfed:
Xcode 16 allows for plugins to determine their own display version, which enables us to:
1. Write a package command plugin that can download the current release build of the `SafeDITool` from Github
2. Use that release version if it exists in our package build plugin

This PR does both of these things. Note that we do not touch the Swift 5.10 & Xcode 15 code-path, which still looks for an unversioned :sob: release installed via Homebrew.

### DIFF
--- a/Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
+++ b/Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
@@ -1,0 +1,91 @@
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import PackagePlugin
+
+@main
+struct InstallSafeDITool: CommandPlugin {
+    func performCommand(
+        context: PackagePlugin.PluginContext,
+        arguments _: [String]
+    ) async throws {
+        guard let safeDIOrigin = context.package.dependencies.first(where: { $0.package.displayName == "SafeDI" })?.package.origin else {
+            print("No package origin found for SafeDI package")
+            return
+        }
+        switch safeDIOrigin {
+        case let .repository(url, displayVersion, _):
+            let expectedToolFolder = context.package.directoryURL.appending(
+                components: ".safedi",
+                displayVersion
+            )
+            let expectedToolLocation = expectedToolFolder.appending(component: "safeditool")
+
+            guard let url = URL(string: url)?.deletingPathExtension() else {
+                print("No package origin found for SafeDI package")
+                return
+            }
+            #if arch(arm64)
+                let toolName = "SafeDITool-arm64"
+            #elseif arch(x86_64)
+                let toolName = "SafeDITool-x86_64"
+            #else
+                print("Unexpected architecture type")
+                return
+            #endif
+
+            let downloadURL = url.appending(
+                components: "releases",
+                "download",
+                displayVersion,
+                toolName
+            )
+            let (downloadedURL, _) = try await URLSession.shared.download(
+                for: URLRequest(url: downloadURL)
+            )
+
+            try FileManager.default.createDirectory(
+                at: expectedToolFolder,
+                withIntermediateDirectories: true
+            )
+            try FileManager.default.moveItem(
+                at: downloadedURL,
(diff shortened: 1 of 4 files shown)


Plugins/SafeDIGenerator/SafeDIGenerateDependencyTree.swift
```diff
@@ -1,3 +1,23 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
dfed, line 19: oops. this header should always have been here

Plugins/SafeDIGenerator/SafeDIGenerateDependencyTree.swift
```diff
@@ -39,16 +59,32 @@ struct SafeDIGenerateDependencyTree: BuildToolPlugin {
...
+            if context.hasSafeDIFolder, downloadedToolLocation == nil {
+                Diagnostics.error("""
+                \(context.safediFolder.path()) exists, but there is no SafeDITool binary for this release present.
+
+                To download the current release SafeDITool binary, run:
+                \tswift package --package-path \(context.package.directoryURL.path()) --allow-network-connections all --allow-writing-to-package-directory safedi-release-install
+
+                To use a debug SafeDITool binary instead, remove the `.safedi` directory by running:
+                \trm -rf \(context.safediFolder.path())
+                """)
```
dfed, line 72: what the error looks like in practice:
IMG_9199

Plugins/SafeDIGenerator/SafeDIGenerateDependencyTree.swift
```diff
@@ -39,16 +59,32 @@ struct SafeDIGenerateDependencyTree: BuildToolPlugin {
...
+            } else if downloadedToolLocation == nil {
+                Diagnostics.warning("""
+                Using a debug SafeDITool binary, which is 15x slower than a release SafeDITool binary.
+
+                To download the current release SafeDITool binary, run:
+                \tswift package --package-path \(context.package.directoryURL.path()) --allow-network-connections all --allow-writing-to-package-directory safedi-release-install
+                """)
```
dfed, line 79: Warning shows up nicely too:
IMG_6124

Plugins/SafeDIGenerator/SafeDIGenerateDependencyTree.swift
dfed: If there is a `.safedi/` directory at the package root, I'm taking that to mean that the user intended to be using a release version of `SafeDITool`. Users will likely enter this state after installing `SafeDITool`, then bumping the `SafeDI` version they are using.

Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
```diff
@@ -0,0 +1,109 @@
...
+        context: PackagePlugin.PluginContext,
+        arguments _: [String]
+    ) async throws {
+        guard let safeDIOrigin = context.package.dependencies.first(where: { $0.package.displayName == "SafeDI" })?.package.origin else {
+            Diagnostics.error("No package origin found for SafeDI package")
+            return
+        }
+        switch safeDIOrigin {
+        case let .repository(url, displayVersion, _):
+            // As of Xcode 16.0 Beta 6, the display version is of the form "Optional(version)".
```
dfed, line 36: Apple... why...

Using `displayVersion` is clearly brittle, but it does work.

MrAdamBoyd: Watch them add a space

Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
```diff
@@ -0,0 +1,110 @@
...
+        guard let safeDIOrigin = context.package.dependencies.first(where: { $0.package.displayName == "SafeDI" })?.package.origin else {
+            Diagnostics.error("No package origin found for SafeDI package")
+            exit(1)
+        }
+        switch safeDIOrigin {
+        case let .repository(url, displayVersion, _):
+            // As of Xcode 16.0 Beta 6, the display version is of the form "Optional(version)".
+            // This regular expression is duplicated by SafeDIGenerateDependencyTree since plugins can not share code.
+            guard let versionMatch = try /Optional\((.*?)\)|^(.*?)$/.firstMatch(in: displayVersion),
+                  let versionSubstring = versionMatch.output.1 ?? versionMatch.output.2
```
MrAdamBoyd, line 39: Why checking `.1` as well as `.2`?

MrAdamBoyd: Ditto in the copied code in the other plugin

Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
dfed: Seemed like an oversight. We'll see https://github.com/swiftlang/swift-package-manager/pull/7934

MrAdamBoyd (approved):
Question but LGTM

Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
dfed: I'm falling back to a version with no surrounding `Optional(...)` in case this gets fixed. `.1` gets us the optional match. `.2` gets us the match without the optional.

Plugins/InstallSafeDITool/InstallCLIPluginCommand.swift
dfed: If this ever fully breaks we can do what Apollo does and [hardcode the current release version](https://github.com/apollographql/apollo-ios/blob/f0c89da06e252c91920fe1e26956b14ba1b17d0a/Sources/Apollo/Constants.swift#L4). I'd prefer to trust SPM rather than have to keep this in-sync myself, so we're starting out with this approach.

PR #123, opened 2024-12-26
Support nested `@Instantiable` types

### LINKED ISSUE #120: Support nested `@Instantiable` types
**Goals**
* Remove [this section](https://github.com/dfed/SafeDI/blob/main/README.md#nested-types) from the README
* Support making more types injectable
* Make it easy to determine how to reference an injectable type

**Non-Goals**
* Make it possible to to reference an injectable type by its non-fully-qualified name

**Investigations**
I prevented SafeDI from including nested types from the outset because of following case:

```swift
@Instantiable
public final class Outer: Instantiable {
    public init(inner: Inner) {
        self.inner = inner
    }

    @Instantiated private let inner: Inner // SafeDITool build error: No Instantiable type "Inner" found.

    @Instantiable
    public final class Inner: Instantiable {
        public init() {}
    }
}
```

It would be unclear to adoptees that `Inner` wasn't found because SafeDI was expecting `let inner: Outer.Inner` instead of `let inner: Inner`. However, now that we've built a ["did you mean"](https://github.com/dfed/SafeDI/pull/108) suggestions into our errors, we've mitigated this error case.

**Design**

We'll need to update `InstantiableVisitor` to enable visible child decls + vending child `Instantiable` types up to the parent. As long as the `Instantiable`'s `instantiableType` type description is the correct nested type, we should be good to go.

It should also be possible to enable referencing nested types by their less-than-fully-qualified names, but that's a larger lift that isn't required for this feature to land. It'd require us to fully qualify types as we create `Instantiable`'s `dependencies` array, which is doable: it'd require our `InstantiableVisitor` to:
1. gather all nested type name information
2. take a pass in a `visitPostDecl(_ node: some ConcreteDeclSyntaxProtocol)` method to update the dependencies to their fully qualified names

dfed:
Resolves #120


README.md
```diff
@@ -389,10 +389,6 @@ The `@StateObject` documentation reads:
 
 In a manual DI system it is simple for superclasses to receive injected dependencies. SafeDI‘s utilization of macros means that SafeDI is not aware of dependencies required due to inheritance trees. Due to this limitation, superclass types should not be decorated with `@Instantiable`: instead, subclasses should declare the properties their superclasses need, and pass them upwards via a call to `super.init(…)`.
 
-#### Nested types
-
-While manually written dependency injection code can work seamlessly with [nested types](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/nestedtypes/), SafeDI currently prevents nested types from being decorated with the `@Instantiable` macro.
```
dfed, line 394: :wave:

Sources/SafeDICore/Models/ConcreteDeclSyntaxProtocol.swift
```diff
@@ -55,6 +55,19 @@ extension ClassDeclSyntax: ConcreteDeclSyntaxProtocol {
...
+        get {
+            enumKeyword
+        }
+        set {
+            enumKeyword = newValue
+        }
+    }
+
+    public var declType: ConcreteDeclType { .enumType }
+}
```
dfed, line 69: We now enable visiting `EnumDeclSyntax` types, because `@Instantiable` types may be declared within them.

Sources/SafeDICore/Visitors/FileVisitor.swift
```diff
@@ -79,58 +79,67 @@ public final class FileVisitor: SyntaxVisitor {
...
         for instantiable in instantiableVisitor.instantiables {
             instantiables.append(instantiable)
         }
 
-        return .skipChildren
+        // Extensions are always top-level.
+        parentType = node.extendedType.typeDescription
+
+        // Continue to find child types.
+        return .visitChildren
```
dfed, line 106: TODO: add a test for finding child types within an extension

Sources/SafeDIMacros/Macros/InjectableMacro.swift
```diff
@@ -141,7 +141,7 @@ public struct InjectableMacro: PeerMacro {
             case .fulfilledByTypeArgumentInvalidType:
                 "The argument `fulfilledByType` must be a string literal"
             case .fulfilledByTypeArgumentInvalidTypeDescription:
-                "The argument `fulfilledByType` must refer to a simple, unnested type"
+                "The argument `fulfilledByType` must refer to a simple type"
```
dfed, line 144: I don't love the term "simple" here but I'm struggling to come up with better terminology.

Sources/SafeDICore/Models/Property.swift
```diff
@@ -25,7 +25,7 @@ import SwiftSyntax
 public struct Property: Codable, Hashable, Comparable, Sendable {
     // MARK: Initialization
 
-    init(
+    public init(
```
dfed, line 28: Maybe instead we want a `withUpdatedTypeDescription` method?

dfed:
Merging this. Going to be a little bit before I release so we have time. If anyone wants to give this a review + has feedback, I will happily address it post-merge.

PR #126, opened 2024-12-28
Explicitly declare root instantiables

dfed:
Finding the `@Instantiable`-decorated types that are roots of a dependency tree automatically is a bit magical, but also comes with a concrete downside: when writing a brand new type that only has `@Instantiated` dependencies, `SafeDI` complains during builds until that new type is `@Instantiated` by another `@Instantiable`-decorated type.

This case is difficult for newcomers to SafeDI to debug, and also annoys veteran SafeDI consumers.

This PR represents a breaking change from the `0.9.*` release, and is intended to land prior to a `1.0.0` release. This PR will land _after_ a `0.9.1` release is created.


NickEntin (approved):
Documentation and tests LGTM

PR #137, opened 2025-01-02
Add Cocoapods support

dfed:
Created a `SafeDI` podspec inspired by [these instructions](https://soumyamahunt.medium.com/support-swift-macros-with-cocoapods-3911f9317042).

In order for the CocoaPods build to not have warnings due to https://github.com/michaeleisel/ZippyJSON/pull/67 not being released yet, I removed `ZippyJSON` from the CocoaPods build by creating a new `Package.swift` file just for CocoaPods that does not have this dependency. This change has no effect on the build runtime for CocoaPods installs, since `ZippyJSON` is only used by `SafeDITool`, not the `SafeDI` framework.

Once `ZippyJSON` releases, I should be able to delete the `CocoaPods` directory entirely and use a single `Package.swift` file again.


README.md
```diff
@@ -82,6 +84,14 @@ dependencies: [
 
 To install the SafeDI framework into an Xcode project with Swift Package Manager, follow [Apple’s instructions](https://developer.apple.com/documentation/xcode/adding-package-dependencies-to-your-app) to add `https://github.com/dfed/SafeDI.git` as a dependency.
 
+#### Cocoapods
```
dfed, line 87:
```suggestion
#### CocoaPods
```

PR #139, opened 2025-01-07
Improve error messages

dfed:
Some better formatting will make the errors easier to read. Right now our error messages are a bit too dense, which makes it easy to miss key info.


Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -2061,7 +2061,7 @@ import SafeDICore
                 @Instantiable
                 public struct ExampleService: Instantiable {
                 public init() {
-                // If the following properties were decorated with the @Instantiated, @Received, or @Forwarded macros, had default values, or were written as computed properties, this initializer could have been auto-generated by the @Instantiable macro.
+                // The following properties are not decorated with the @Instantiated, @Received, or @Forwarded macros, do not have default values, and are not computed properties.
```
dfed, line 2064: Oops. This comment used to explain why SafeDI couldn't auto-generate the initializer for you, back when we used to do that.

Now I'm making it explain why we're leaving you fill-in-the-blanks. I'm not convinced the comment is worth keeping though. Thoughts?

NickEntin: I'd leave it for now. Easy to delete the comment, and really easy to miss this sort of mistake.

NickEntin (approved):

Issue #146, opened 2025-01-14
Document how to adopt SafeDI in UIKit / AppDelegate apps


pouyayarandi:
**Steps to reproduce**
1. Make AppDelegate as your root instantiable
2. Generate SafeDI.swift

**Sample code or sample project upload**
Sample AppDelegate
```swift
import UIKit
import SafeDI
import UserRepo
import Networking

@Instantiable(isRoot: true)
@main
public class AppDelegate: UIResponder, UIApplicationDelegate, Instantiable {

    @Instantiated let viewControllerBuilder: Instantiator<ViewController>
    @Instantiated private let userRepository: UserRepository
    @Instantiated private let network: NetworkService

    public init(
        viewControllerBuilder: Instantiator<ViewController>,
        userRepository: UserRepository,
        network: NetworkService
    ) {
        self.viewControllerBuilder = viewControllerBuilder
        self.userRepository = userRepository
        self.network = network
    }

    public var window: UIWindow?

    public func application(
        _ application: UIApplication,
        didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
    ) -> Bool {
        window = .init(frame: UIScreen.main.bounds)
        window?.rootViewController = viewControllerBuilder.instantiate()
        window?.makeKeyAndVisible()
        return true
    }
}
```

**Expected behavior**
It compiles successfully.

**Actual behavior**
It has two issues:
1. In generated code `override` is missing.
```swift
extension AppDelegate {
    public convenience init() { // compile error: Overriding declaration requires an 'override' keyword
        func __safeDI_network() -> NetworkServiceImpl {
            let session: URLSession = URLSession.instantiate()
            let decoder: JSONDecoder = JSONDecoder.instantiate()
            return NetworkServiceImpl(session: session, decoder: decoder)
        }
        let network: NetworkService = __safeDI_network()
        func __safeDI_viewControllerBuilder() -> ViewController {
            let userRepository: UserRepository = UserRepositoryImpl(network: network)
            return ViewController(userRepository: userRepository)
        }
        let viewControllerBuilder = Instantiator<ViewController>(__safeDI_viewControllerBuilder)
        let userRepository: UserRepository = UserRepositoryImpl(network: network)
        self.init(viewControllerBuilder: viewControllerBuilder, userRepository: userRepository, network: network)
    }
}
```
2. Xcode complains as `init()` is already overriden.
[...]

PR #148, opened 2025-01-16
Document recommendations for root object selection

dfed:
Addresses #146

### DIFF
--- a/Documentation/Manual.md
+++ b/Documentation/Manual.md
@@ -404,7 +404,15 @@ In a manual DI system it is simple for superclasses to receive injected dependen
 
 It is strongly recommended that projects adopting SafeDI start their migration by identifying the root of their dependency tree and making it `@Instantiable(isRoot: true)`. Once your root object has adopted SafeDI, continue migrating dependencies to SafeDI in either a breadth-first or depth-first manner. As your adoption of SafeDI progresses, you’ll find that you are removing more code than you are adding: many of your dependencies are likely being passed through intermediary objects that do not utilize the dependency except to instantiate a dependency deeper in the tree. Once types further down the dependency tree have adopted SafeDI, you will be able to avoid receiving dependencies in intermediary types.
 
-## Example Application
+### Selecting a root in SwiftUI applications
+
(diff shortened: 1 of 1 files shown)


pouyayarandi:
Great job @dfed 
Does it mean that we should store SafeDI.swift file somewhere inside project? We should call `Root.init()` inside `AppDelegate` and without storing generated code, Xcode must fail compilation as `Root.init()` does not exist.
I'd prefer not to store this file because of high probability of code conflict it brings.

dfed:
@pouyayarandi if you're using a build plugin the generated tree is in derived data. If you're using a pre-build script and generating the file into your git repo, I would recommend ignoring the file into `git`

Documentation/Manual.md
```diff
@@ -404,7 +404,15 @@ In a manual DI system it is simple for superclasses to receive injected dependen
...
+
+SwiftUI applications have a natural root: the `App`-conforming type that is initialized when the binary is launched.
+
+### Selecting a root in UIKit applications
+
+UIKit applications’ natural root is the `UIApplicationDelegate`-conforming app delegate, however, this type inherits from the Objective-C `NSObject` which already has a no-argument `init()`. As such, it is best to create a custom `@Instantiable(isRoot: true) public final class Root: Instantiable` type that is initialized and stored by the application’s app delegate.
+
+## Example application
 
 We’ve tied everything together with an example multi-user notes application backed by SwiftUI. You can compile and run this code in Xcode in the included [ExampleProjectIntegration](../Examples/ExampleProjectIntegration) project.
```
dfed, line 417:
```suggestion
We’ve tied everything together with an example multi-user notes application backed by SwiftUI. You can compile and run this code in Xcode in the included [ExampleProjectIntegration](../Examples/ExampleProjectIntegration) project. This same multi-user notes app also exists in [an example Xcode project with multiple modules](../Examples/ExampleMultiProjectIntegration), and also in [an example Xcode project using CocoaPods](../Examples/ExampleCocoaPodsIntegration). We have also created [an example multi-module `Package.swift` that integrates with SafeDI](../Examples/ExamplePackageIntegration).
```

PR #158, opened 2025-05-23
Run Xcode tests on Xcode 16.1

dfed:


dfed:
Nope. That's not it! Looks like I've got a test to debug. Not sure what changed in the migration to Swift Testing to make this test unreliable. It always succeeds when running the test on its own. So likely a global state issue.

PR #159, opened 2025-05-24
Improve test reliability

dfed:
The move to Swift Testing in #156 introduced some nondeterminism due to our use of a single static pointer `FileFinder`. Swift Testing runs multiple tests in parallel, and multiple tests were smashing the `FileFinder` pointer.

This PR does two things:
1. Uses [TaskLocal](https://developer.apple.com/documentation/swift/tasklocal) to eliminate the possibility each test uses the desired `FileFinder`.
2. Enables all Swift Test suites to run in parallel.


Tests/SafeDIToolTests/SafeDIToolCodeGenerationErrorTests.swift
```diff
@@ -24,8 +24,7 @@ import Testing
 
 @testable import SafeDITool
 
-@Suite(.serialized)
-final class SafeDIToolCodeGenerationErrorTests {
+struct SafeDIToolCodeGenerationErrorTests: ~Copyable {
```
dfed, line 27: The `~Copyable` conformance is required for `deinit` to work on a `struct

Tests/SafeDIToolTests/SafeDIToolCodeGenerationErrorTests.swift
```diff
@@ -40,8 +39,8 @@ final class SafeDIToolCodeGenerationErrorTests {
 
 	// MARK: Error Tests
 
-	@MainActor @Test
-	func run_onCodeWithPropertyWithUnknownFulfilledType_throwsError() async {
+	@Test
+	mutating func run_onCodeWithPropertyWithUnknownFulfilledType_throwsError() async {
```
dfed, line 43: Because we pass in a modifiable `filesToDelete: &filesToDelete` in each `executeSafeDIToolTest` call, we need the test method to be mutating.

Tests/SafeDIToolTests/SafeDIToolCodeGenerationErrorTests.swift
```diff
@@ -1795,27 +1794,25 @@ final class SafeDIToolCodeGenerationErrorTests {
...
+			tool.dotFileOutput = nil
+			await assertThrowsError("Could not create file enumerator for directory 'Fake'") {
+				try await tool.run()
+			}
 		}
 	}
 
-	func test_include_throwsErrorWhenNoSwiftSourcesFilePathAndNoInclude() async {
+	@Test
+	func include_throwsErrorWhenNoSwiftSourcesFilePathAndNoInclude() async {
```
dfed, line 1815: eek. Missed converting this test in #156. Found a couple like this.

.github/workflows/ci.yml
```diff
@@ -144,12 +144,12 @@ jobs:
           sudo xcodebuild -downloadPlatform visionOS
           sudo xcodebuild -runFirstLaunch
       - name: Lint Podspec
-        run: bundle exec pod lib lint --verbose --fail-fast --swift-version=6.0 --platforms=${{ matrix.platforms }}
```
dfed, line 147: this version comes from the `.swiftformat` file – duplicating here is a recipe for a future failed build

.github/workflows/ci.yml
```diff
@@ -144,12 +144,12 @@ jobs:
...
+        run: bundle exec pod lib lint --verbose --fail-fast --platforms=${{ matrix.platforms }}
 
   linux:
     name: Build and Test on Linux
     runs-on: ubuntu-latest
-    container: swift:6.0
+    container: swift:6.1
```
dfed, line 152: Swift Testing on Swift 6 doesn't like using `~Copyable` structs in tests

.github/workflows/ci.yml
```diff
@@ -178,7 +178,7 @@ jobs:
   lint-swift:
     name: Lint Swift
     runs-on: ubuntu-latest
-    container: swift:6.0
+    container: swift:6.1
```
dfed, line 181: Not a functional change, but doesn't seem harmful

CLI/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 6.0
+// swift-tools-version: 6.1
```
dfed, line 1: development in this lib should be in Swift 6.1. We just need to support building in Swift 6.

.github/workflows/ci.yml
dfed: oops. this is cocoa pods! not the linter!

PR #164, opened 2025-06-01
Enable FixIt to update existing initializer rather than always providing a new one

dfed:
When an `@Instantiable` does not have a properly formatted `init`, and one of the type's `init` methods only is missing arguments or has incorrect/missing accessibility modifiers, our FixIt now updates an existing `init` method rather than creating a new one.

### DIFF
--- a/Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
+++ b/Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
@@ -2219,7 +2219,7 @@ import Testing
 				"""
 				@Instantiable
 				public final class UserService: Instantiable {
-					final public init(a: A, b: B, c: C) {
+					public final init(a: A, b: B, c: C) {
 						self.a = a
 						self.b = b
 						self.c = c
@@ -2233,7 +2233,7 @@ import Testing
 			} expansion: {
 				"""
 				public final class UserService: Instantiable {
-					final public init(a: A, b: B, c: C) {
+					public final init(a: A, b: B, c: C) {
 						self.a = a
 						self.b = b
 						self.c = c
--- a/Sources/SafeDIMacros/Macros/InstantiableMacro.swift
+++ b/Sources/SafeDIMacros/Macros/InstantiableMacro.swift
@@ -264,17 +264,20 @@ public struct InstantiableMacro: MemberMacro {
 							}
 						})
 						if !fixedSyntax.modifiers.containsPublicOrOpen {
-							fixedSyntax.modifiers.append(.init(
+							if syntaxToFix.modifiers.first != nil {
+								fixedSyntax.modifiers[fixedSyntax.modifiers.startIndex].leadingTrivia = []
+							}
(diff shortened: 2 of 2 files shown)


Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -1196,68 +1330,594 @@ import Testing
...
-				self.receivedA = receivedA
-				self.receivedB = receivedB
-				}
-
-				    public init(forwardedA: ForwardedA, receivedA: ReceivedA) {
-				        self.forwardedA = forwardedA
-				        self.receivedA = receivedA
-				        receivedB = ReceivedB()
-				    }
+					init(forwardedA: ForwardedA, receivedA: ReceivedA, receivedB: ReceivedB, customizable: String = "") {
```
NickEntin, line 1912: Should it also add `public` here?

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -1453,6 +2113,534 @@ import Testing
...
+					@Received let a: A
+					@Instantiated let b: B
+					@Forwarded let c: C
+				}
+				"""
+			} fixes: {
+				"""
+				@Instantiable
+				public final class UserService: Instantiable {
+					final public init(a: A, b: B, c: C) {
```
NickEntin, line 2222: nit: `public final`? linter will catch it too, so doesn't really matter

NickEntin: (and see the line above)

NickEntin (approved):
Tests LGTM

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
NickEntin: I guess will just do two passes on fix-its right now

dfed: Ideally yes, but today we just does one fix at a time. There will be new fixit once the first fixit is applied.

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
dfed: 🤔 I can prepend instead of postpend since that's likely what folk will want. will play.

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
dfed: you beat me to it 😉

Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
dfed: addressed in 373e4b3dc0ec632626060d6ee9b035278604c458

PR #167, opened 2025-06-03
Only write the dependency tree output file if it has changed

dfed:
@ahmdmhasn had a great idea to not touch the output file if the output file's content hasn't changed. Looks like modern Xcode seems to support having a package plugin that does supports this, so let's do it.

I believe earlier SPM implementations required the output file to be touched as part of the build plugin run, so we _needed_ to write the file. This does not seem to be the case in Xcode 16+.

### DIFF
--- a/Sources/SafeDITool/SafeDITool.swift
+++ b/Sources/SafeDITool/SafeDITool.swift
@@ -52,6 +52,13 @@ struct SafeDITool: AsyncParsableCommand, Sendable {
 			throw ValidationError("Must provide 'swift-sources-file-path', '--include', or '--include-file-path'.")
 		}
 
+		async let existingGeneratedCode: String? = Task.detached {
+			if let dependencyTreeOutput {
+				try? String(contentsOf: dependencyTreeOutput.asFileURL, encoding: .utf8)
+			} else {
+				nil
+			}
+		}.value
 		let (dependentModuleInfo, module) = try await (
 			loadSafeDIModuleInfo(),
 			parsedModule()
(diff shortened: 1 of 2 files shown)


Sources/SafeDITool/SafeDITool.swift
```diff
@@ -52,6 +52,13 @@ struct SafeDITool: AsyncParsableCommand, Sendable {
 			throw ValidationError("Must provide 'swift-sources-file-path', '--include', or '--include-file-path'.")
 		}
 
+		async let existingGeneratedCode: String? = Task.detached {
+			if let dependencyTreeOutput {
+				try? String(contentsOf: dependencyTreeOutput.asFileURL, encoding: .utf8)
```
ahmdmhasn, line 57: Do we need to log if unable to initialize a string with the content of the tree output? Anyway, it should be `nil` for initialization errors, or the file doesn't exist initially.

Sources/SafeDITool/SafeDITool.swift
```diff
@@ -52,6 +52,13 @@ struct SafeDITool: AsyncParsableCommand, Sendable {
...
+			} else {
+				nil
+			}
+		}.value
 		let (dependentModuleInfo, module) = try await (
```
ahmdmhasn, line 62:
```suggestion
		}.value

		let (dependentModuleInfo, module) = try await (
```
nit; Extra space after `existingGeneratedCode` computation.

Sources/SafeDITool/SafeDITool.swift
```diff
@@ -124,7 +131,11 @@ struct SafeDITool: AsyncParsableCommand, Sendable {
 			)).write(toPath: moduleInfoOutput)
 		}
 
-		if let dependencyTreeOutput, let generatedCode = try await generatedCode {
+		if let dependencyTreeOutput,
+		   let generatedCode = try await generatedCode,
+		   // Only update the file if the file has changed.
```
ahmdmhasn, line 136: Nice inline comment

ahmdmhasn (approved):
Good job. I hope this makes build time even faster for incremental builds. 🙌🏼

Sources/SafeDITool/SafeDITool.swift
dfed: I don't think so – if we can't read it for whatever reason we'll just write a new one.

PR #169, opened 2025-06-29
Initializer fixit should not re-add conforming closure parameter

dfed:
#164 introduced an issue where a fixit to add new parameters to an initializer would repeat any closure parameters in the list. This PR addresses that issue.

### DIFF
--- a/Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
+++ b/Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
@@ -1860,6 +1860,80 @@ import Testing
 		}
 
 		@Test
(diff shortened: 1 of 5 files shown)


Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -756,7 +756,7 @@ import Testing
 				    public init(closure: @escaping @Sendable () -> Void) {
 				        self.closure = closure
 				    }
-				    @Forwarded let closure: (() -> Void)
```
dfed, line 759: found a bug while simplifying this code!

PR #182, opened 2025-09-17
Bump swift-syntax to 602.0.0

dfed:


dfed:
nevermind. This'll require updating CI to Xcode 26. We'll get there later.

PR #190, opened 2026-01-16
Improve initializer fixit when there are no parameters and no body

dfed:
We should ensure assignments are on their own line!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,6 @@ jobs:
           sudo xcodebuild -runFirstLaunch
           sudo xcrun simctl list
           sudo xcodebuild -downloadPlatform visionOS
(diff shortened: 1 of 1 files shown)


Tests/SafeDIMacrosTests/InstantiableMacroTests.swift
```diff
@@ -1935,6 +1935,76 @@ import SafeDICore
...
+					@Instantiated let instantiated: Instantiated
+					@Forwarded let forwarded: Forwarded
+				}
+				"""
+			} fixes: {
+				"""
+				@Instantiable
+				public struct ExampleService: Instantiable {
+					public init(
+				received: Received,
```
dfed: Xcode will handle adding the correct indentation

PR #236, opened 2026-04-10
Move onlyIfAvailable received dependencies into SafeDIParameters (simple)

dfed:
## Summary
- They become plain optional values (`Type? = nil`) on `SafeDIParameters`
- When `nil`, the dependency is absent; when provided (e.g., `.mock()`), the value is used directly
- Fixed a duplicate-promotion bug when both a required `@Received` and an `onlyIfAvailable` `@Received` existed for the same label+type

**Alternative:** See #237 for a tree-threading version that uses `SafeDIMockConfiguration?` instead of plain optional values, enabling subtree override through config structs.

## Test plan
- [x] All 744 tests pass
- [x] SwiftFormat lint passes
- [x] Verified against apple repo integration


dfed:
Superseded by new PR with rebased branch

PR #237, opened 2026-04-10
Move onlyIfAvailable received dependencies into SafeDIParameters (tree-threading)

dfed:
## Summary
- They become optional config structs (`Type.SafeDIMockConfiguration? = nil`) on `SafeDIParameters`
- When `nil`, the dependency is absent; when `.init()`, the full subtree is built with tree-threading support — including transitive `@Received` dependencies
- Uses `suppressReceivedPropagation` to prevent transitive deps from leaking to the root scope
- Enriched scopes include `requiredReceivedProperties` as children so they appear in the config struct
- Config struct dedup prefers the enriched version (more children) when the same type appears in both regular and onlyIfAvailable contexts

**Alternative:** See #236 for a simpler version that uses plain `Type? = nil` without tree-threading.

## Test plan
- [x] All 745 tests pass (includes dedup coverage test)
- [x] SwiftFormat lint passes
- [ ] Verify against apple repo integration


dfed:
Superseded by new PR with rebased branch

PR #238, opened 2026-04-10
Revert leaf mock wrapper generation

dfed:
## Summary
- revert the always-wrapper mock generation change for simple leaf nodes
- restore the pre-alpha-14 mock API shape where simple leaves stay raw builder values and richer nodes use `SafeDIMockConfiguration`
- keep the unrelated launch-only SafeDITool fallback for plugin setup intact

## Why
The always-wrapper change appears to have caused a meaningful compile-time regression in the Apple repo after the `2.0.0-alpha-14` bump. Reverting just that generated API change lets us preserve the launch fallback fix while removing the extra mock-wrapper surface area.

## What changed
- restored `needsConfigurationStruct`-based branching in mock generation
- restored raw optional leaf builder emission in root `SafeDIParameters` and child configs
- restored the previous shared-config ownership behavior for simple leaf nodes
- updated mock-generation golden tests back to the pre-wrapper shape
- left `Plugins/Shared.swift` and `Plugins/SafeDIGenerator/SafeDIGenerateDependencyTree.swift` unchanged so the local source fallback still only triggers on tool launch failures

## Validation
- `swift build --traits sourceBuild`
- `swift test --traits sourceBuild`
- `./CLI/lint.sh`


README.md
```diff
@@ -121,6 +121,8 @@ To also generate mocks for non-root modules, add the plugin to all first-party t
 
 You can see this integration in practice in the [Example Package Integration](Examples/Example Package Integration) package.
 
+Types annotated with `@Instantiable(generateMock: true)` get generated `mock()` methods for testing and SwiftUI previews. When a mock root has an `@Instantiated` dependency subtree, its generated `mock()` accepts a `safeDIParameters` argument for tree-structured overrides. Simple leaf overrides remain direct builder values, while richer child nodes use `SafeDIMockConfiguration`.
```
dfed:
```suggestion
```

README.md
```diff
@@ -121,6 +121,8 @@ To also generate mocks for non-root modules, add the plugin to all first-party t
...
+
```
dfed:
```suggestion
```

PR #242, opened 2026-04-10
Move onlyIfAvailable received dependencies into SafeDIParameters (tree-threading)

dfed:
## Summary
- They become optional config structs (`Type.SafeDIMockConfiguration? = nil`) or optional closures on `SafeDIParameters`
- When `nil`, the dependency is absent; when `.init()`, the full subtree is built with tree-threading support
- Uses `suppressReceivedPropagation` to prevent transitive deps from leaking to the root scope
- Enriched scopes include `requiredReceivedProperties` as children so they appear in the config struct
- Config struct dedup prefers the enriched version when the same type appears in both regular and onlyIfAvailable contexts

**Alternative:** See #241 for a simpler version that uses plain `Type? = nil` without tree-threading.

## Test plan
- [x] All 743 tests pass (includes dedup coverage test)
- [x] SwiftFormat lint passes
- [x] Code coverage verified
- [ ] Verify against apple repo integration


dfed:
Tree-threading for onlyIfAvailable can't work with shared config structs: received deps without scopes (String, URL, etc.) can't be enriched, and propagating them to root causes the same leaking problem this change aims to fix. The simple approach (#241) is the right answer.

PR #248, opened 2026-04-11
Add mockOnly parameter to @Instantiable

dfed:
## Summary

- Adds `mockOnly: Bool = false` to `@Instantiable`, allowing types to provide a hand-written `mock()` method without requiring `init`/`instantiate()` or `Instantiable` conformance
- When a dependency's type has a `mockOnly` declaration, the mock generator uses `Type.mock()` as the default — forwarded deps become defaulted parameters, and instantiated deps become optional tree children
- Supports `customMockName` for naming the hand-written method
- Combines mockOnly + non-mockOnly `@Instantiable` declarations for the same type via `mergedWithMockProvider(_:)` (e.g., production init from one, mock from the other)
- Works on extensions of stdlib/Foundation types (e.g., `@Instantiable(mockOnly: true) extension String`)
- Raises codecov target from 99.9% to 100%

## Test plan

- [x] 12 macro expansion tests (success cases, mutual exclusion errors, missing mock method errors, customMockName interactions, extension branch coverage)
- [x] 10 code generation tests (forwarded defaults, instantiated tree children, primitive types, custom mock names, same-type merge permutations, production isolation)
- [x] 5 error tests (duplicate mock providers — all ordering permutations)
- [x] 6 FixableInstantiableError unit tests (description + fixIt for each new case)
- [x] 2 production isolation tests (verify production output is unchanged when mockOnly is added)
- [x] All 777 tests pass, lint clean


Documentation/Manual.md
```diff
@@ -538,10 +538,65 @@ public struct NetworkClient: Instantiable {
...
 ```
 
-The `customMockName` parameter requires `generateMock: true`.
+The `customMockName` parameter requires `generateMock: true` or `mockOnly: true`.
 
 If you provide a mock method without `generateMock: true`, parent types that instantiate the child will call `ChildType.mock(...)` (or `ChildType.customMock(...)`) instead of `ChildType(...)` when constructing it, threading mock parameters through your custom method.
 
+### The `mockOnly` parameter
+
+The `mockOnly` parameter lets you provide a hand-written `mock()` method for types that don't need full `@Instantiable` infrastructure. Here's an example providing a mock for a third-party type:
```
dfed:
```suggestion
The `mockOnly` parameter lets you provide a hand-written `mock()` method for types that don’t need full `@Instantiable` infrastructure. Here’s an example providing a mock for a third-party type:
```

Documentation/Manual.md
```diff
@@ -538,10 +538,65 @@ public struct NetworkClient: Instantiable {
...
+This also works for primitive types used as `@Forwarded` dependencies:
+
+```swift
+@Instantiable(mockOnly: true)
+extension String {
+    public static func mock() -> String { "" }
+}
+```
+
+When a parent type references a `mockOnly` type as a dependency, the generated mock uses `Type.mock()` as the default. For `@Forwarded` dependencies, the parameter gets a default value so callers don't need to provide it. For `@Instantiated` dependencies, the type appears in `SafeDIOverrides` with `Type.mock()` as the default, allowing optional override.
```
dfed:
```suggestion
When a parent type references a `mockOnly` type as a dependency, the generated mock uses `Type.mock()` as the default. For `@Forwarded` dependencies, the parameter gets a default value so callers don’t need to provide it. For `@Instantiated` dependencies, the type appears in `SafeDIOverrides` with `Type.mock()` as the default, allowing optional override.
```

Documentation/Manual.md
```diff
@@ -538,10 +538,65 @@ public struct NetworkClient: Instantiable {
...
+
+`mockOnly` is useful for:
+
+- Types defined in other modules (e.g., third-party dependencies) that need mocks in your tests
+- Primitive or Foundation types used as `@Forwarded` dependencies (e.g., `String`, `Int`, `UUID`)
+- Types whose `@Instantiable` declaration is in another module and isn't in the current module's dependency tree
```
dfed:
```suggestion
- Types whose `@Instantiable` declaration is in another module and isn’t in the current module’s dependency tree
```

Sources/SafeDICore/Generators/ScopeGenerator.swift
```diff
@@ -741,6 +744,9 @@ actor ScopeGenerator: CustomStringConvertible, Sendable {
 			let typeSource = dependency.property.typeDescription.asFunctionParameter.asSource
 			if let defaultExpression = constructionDefaults[dependency.property.label] {
 				mockParameters.append("\(bodyIndent)\(dependency.property.label): \(typeSource) = \(defaultExpression)")
+			} else if let mockMethodName = context.handWrittenMockTypes[dependency.property.typeDescription] {
+				let mockTypeName = dependency.property.typeDescription.asSource
+				mockParameters.append("\(bodyIndent)\(dependency.property.label): \(typeSource) = \(mockTypeName).\(mockMethodName)()")
```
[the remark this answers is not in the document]
dfed: Good catch — this is a real edge case. When a hand-written mock has required dependency-matching parameters, the generated `Type.mock()` default won't compile. Fixed in #249 by tightening the `handWrittenMockTypes` filter to only include types whose mock can be called with zero arguments.

Sources/SafeDIMacros/Macros/InstantiableMacro.swift
```diff
@@ -694,6 +749,44 @@ public struct InstantiableMacro: MemberMacro {
...
+					node: Syntax(node),
+					error: FixableInstantiableError.mockOnlyMissingMockMethod(
+						typeName: extensionDeclaration.extendedType.typeDescription.asSource,
+						methodName: expectedMethodName,
+					),
+					changes: Self.generateCustomMockStub(
+						named: expectedMethodName,
+						typeName: extensionDeclaration.extendedType.typeDescription.asSource,
+						dependencies: [],
+						isExtension: true,
```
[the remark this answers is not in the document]
dfed: Not a real concern. The diagnostic at this line *is* the enforcement that a mock method must exist — it fires when the method is missing. The fix-it is just a convenience scaffold; the user has to fill in the body with real construction logic anyway since SafeDI can't know how to build an arbitrary third-party type. The stub body being imperfect is fine.

PR #249, opened 2026-04-12
Fix forwarded mock default when mock has required parameters

dfed:
## Summary

- When a hand-written mock method has required dependency-matching parameters (e.g. `mock(name: String)` where `name` is a `@Received` dependency), the generated default `Type.mock()` for forwarded parameters would not compile
- Tightens the `handWrittenMockTypes` filter in `DependencyTreeGenerator` to only include types whose mock can be called with zero arguments (`allSatisfy(\.hasDefaultValue)`)
- Types with required mock parameters are excluded, so the forwarded parameter stays required instead of getting an invalid default

## Test plan

- [x] Added `mock_omitsDefaultForForwardedParameter_whenHandWrittenMockHasRequiredParameters` test
- [x] Full test suite passes (781 tests)
- [x] Lint passes


dfed:
@codex review

dfed:
@codex review

PR #257, opened 2026-04-18
Fix mock generation using init signature instead of hand-written mock

dfed:
## Summary

When a type declares a hand-written mock with fewer parameters than its
initializer, `collectMockParameterTree` was still treating the init as
the source of construction arguments. If the mock had zero arguments,
the codegen:

1. Nilled out `constructionInitializer` (so default parameters were
   skipped correctly)…
2. …but then fell back to `childInstantiable.initializer?.arguments` for
   `constructionArguments`, re-introducing the init's parameters into
   the generated call.

Result: the generated mock referenced a signature that did not exist on
the hand-written mock, e.g.

```swift
let service = (safeDIOverrides.service ?? Service.customMock(onCancel:onSubmit:))({ _ in }, { _ in })
```

even though `Service.customMock` was declared as `customMock() -> Self`.

## Fix

`collectMockParameterTree` now picks a single `constructionInitializer`
(mock when `useMockInitializer && mockInitializer != nil`, otherwise
init) and derives both `defaultParameters` and `constructionArguments`
from it. The empty-mock case produces empty args and no defaults,
correctly emitting `(safeDIOverrides.service ?? Service.customMock)()`.

## Test plan

- [x] Added `mock_referencesHandWrittenMockSignature_whenInitHasAdditionalDefaultedParams` regression
- [x] Updated existing `mock_typeWithUserMockAndOnlyDefaultValuedParamsGeneratesCallThrough_whenCustomMockNameIsSet` expected output (previously encoded the bug)
- [x] Updated `mock_referencesExpectedMockSignature_whenCustomMockMethodMissingDependencyParameters` expected output (same bug, missing-deps error-path variant)
- [x] `swift test --traits sourceBuild` — all 825 tests pass
- [x] `./CLI/lint.sh`


dfed:
@codex review

PR #266, opened 2026-04-19
Generator: mock() returns customMock's return type when it differs from Self

dfed:
## Summary
- Fix generator to declare `static func mock(...)` with `customMock`'s return type when `customMock` returns an additional type (e.g. a protocol fulfilled via `fulfillingAdditionalTypes`) instead of the concrete type. Previously the method signature returned the concrete type but its body called `customMock(...)` returning the additional type, producing a type error.
- Remove `skipCompileVerification: true` from `mock_extensionUserMockReturningAdditionalTypeUsedForProtocolProperty` and `mock_extensionUserMockReturningAdditionalTypeNotUsedForConcreteProperty`, and add explicit expectations for the extension's own mock file.
- Document the new behavior in the `customMockName` section of `Documentation/Manual.md`.

## Why

FIXMEs at `SafeDIToolMockGenerationCustomMockTests.swift:1480` and `:1565` — both the same root cause. With compile verification now running on every test (#256), these were the last extension-customMock tests still opting out for a real generator bug.

## Fix detail

`mockMethodReturnType` in `generateMockRootCode` now picks `mockReturnType.asSource` when the custom mock's declared return type is NOT compatible with the concrete type (i.e. it's not `Self` / the concrete type / the property type). In every other case it stays on the concrete type name, so existing non-protocol-returning customMock tests are unchanged.

The macro's `mockMethodIncorrectReturnType` check rejects this shape for concrete (struct) types, so the new code path is only reachable for extension-based Instantiables.

## Self-review pass
- Pinned the extension's own mock file on the sibling `NotUsedForConcreteProperty` test so silent drift can't go unnoticed (`083fc16`).
- Added a Manual.md paragraph under `customMockName` explaining when the generated `mock()` returns an additional type (`f760b82`).

## Stacked on #256

Base: `claude/compile-verify-codegen-tests`. Three commits on top: `6a77c78`, `083fc16`, `f760b82`.

## Test plan
- [x] `swift test --traits sourceBuild` — 841 tests pass
- [x] `./CLI/lint.sh`


dfed:
@codex review

PR #279, opened 2026-04-21
Rename mock `safeDIOverrides:` external label to `overrides:`

dfed:
## Summary
- The generator-emitted parameter on `mock()` becomes `overrides safeDIOverrides: SafeDIOverrides = .init()`. Internal name `safeDIOverrides` and the `SafeDIOverrides` struct type are unchanged, so every reference inside the mock body and every generated struct declaration stay as they were.
- `mock()` is already namespaced by SafeDI and `SafeDIOverrides` still carries the namespace at the type level, so the `safeDIOverrides:` call-site label was ceremonial. Shorter label = nicer call sites.
- Breaking change at the call site, but the internal name lets userland `@Forwarded let overrides: ...` coexist — Swift allows duplicate external argument labels when internal names differ. A new disambiguation-suite test (`mock_emitsBothParameters_whenForwardedPropertyIsNamedOverrides`) locks that behavior in.

## Test plan
- [x] `swift test --traits sourceBuild` — 909 tests pass
- [x] `./CLI/lint.sh` — clean
- [x] New test covers a `@Forwarded let overrides: String` alongside an `@Instantiated` dep, asserting both `overrides: String` and `overrides safeDIOverrides: SafeDIOverrides = .init()` appear in the generated signature


dfed:
@codex review

Documentation/Manual.md
```diff
@@ -623,19 +623,19 @@ Your user-defined `mock()` method must be `public` (or `open`) and must accept p
 
 ### Overriding dependencies
 
-When a type has `@Instantiated` dependencies, the generated `mock()` accepts a `safeDIOverrides` argument that lets you override any dependency in the tree. Each entry on `SafeDIOverrides` is either a closure whose parameters match the resolved values of that dependency’s own inputs, or a nested `SafeDIMockConfiguration` struct when the dependency has its own `@Instantiated` subtree or default-valued init parameters.
+When a type has `@Instantiated` dependencies, the generated `mock()` accepts an `overrides` argument that lets you override any dependency in the tree. Each entry on `SafeDIOverrides` is either a closure whose parameters match the resolved values of that dependency’s own inputs, or a nested `SafeDIMockConfiguration` struct when the dependency has its own `@Instantiated` subtree or default-valued init parameters.
```
dfed, line 626:
```suggestion
When a type has `@Instantiated` dependencies, the generated `mock()` accepts an `overrides` argument that lets you override any dependency in the tree. Each entry on these `SafeDIOverrides` is either a closure whose parameters match the resolved values of that dependency’s own inputs, or a nested `SafeDIMockConfiguration` struct when the dependency has its own `@Instantiated` subtree or default-valued `init` parameters.
```

PR #280, opened 2026-04-21
Docs: consolidate closure override examples

dfed:
## Summary
- The \"Overriding dependencies\" section of the manual had two closure-shaped-entry examples that didn't carve nature at a joint — both demonstrated \"closure-shaped entry, filled with a closure,\" just with different closure arities. `{ InMemoryStorage() }` could equally be written `InMemoryStorage.init`, so the arity variance wasn't the teaching moment.
- Collapsed them into one example that shows both the closure-literal form and the `init`-reference shorthand side-by-side, with prose that names the shorthand explicitly.

## Test plan
- [x] Manual re-read — the example now carries both forms in one block and the prose explains when each applies

### DIFF
--- a/Documentation/Manual.md
+++ b/Documentation/Manual.md
@@ -629,7 +629,7 @@ Closure-shaped entries apply when the dependency has nothing further to configur
 
 ```swift
 LoggedInView.mock(safeDIOverrides: .init(
-    // Literal closure — `StringStorage` has no inputs of its own, so its closure takes nothing:
+    // Literal closure — `StringStorage` has no inputs of its own, so its closure has no parameters:
     stringStorage: { InMemoryStorage() },
     // Init reference — `StubUserService.init(stringStorage:)` is already `(StringStorage) -> UserService`:
     userService: StubUserService.init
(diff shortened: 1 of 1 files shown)


Documentation/Manual.md
```diff
@@ -625,17 +625,13 @@ Your user-defined `mock()` method must be `public` (or `open`) and must accept p
...
 ```swift
-// Override a leaf dependency — UserDefaults.instantiate() takes nothing, so its closure takes nothing:
-LoggedInView.mock(safeDIOverrides: .init(
-    stringStorage: { InMemoryStorage() }
-))
-
-// `StubUserService.init(stringStorage:)` already matches the override closure's shape
-// — `(StringStorage) -> UserService` — so the initializer can be passed directly:
 LoggedInView.mock(safeDIOverrides: .init(
+    // Literal closure — `StringStorage` has no inputs of its own, so its closure takes nothing:
```
dfed:
```suggestion
    // Literal closure — `StringStorage` has no inputs of its own, so its closure has no parameters:
```

PR #297, opened 2026-05-26
Fail closed on conditional SafeDI scanning

adam-zethraeus:
SafeDITool walks every branch of a SwiftSyntax conditional-compilation while scanning for SafeDI declarations — potentially leaving inactive declarations and members visible to code generation and allowing unknown target-specific conditions to silently affect the dependency graph.

This PR closes the loophole by adding a conditional-compilation evaluator for the scanner. It selects active #if/#elseif/#else clauses, supports boolean literals plus !, &&, ||, and parentheses, and treats custom flags as active only when supplied through the new active-compilation-conditions input.

If an unevaluable condition guards a SafeDI declaration or relevant instantiable member, parsing fails closed with an actionable error.

Disclosure:
- The implementation was AI assisted.
- I haven't opened an issue to discuss this as a feature rather than a bug fix. I hope that's okay, and am more than happy to discuss if that would be useful.

Thanks!

### DIFF
--- a/Sources/SafeDICore/Models/SafeDIToolManifest.swift
+++ b/Sources/SafeDICore/Models/SafeDIToolManifest.swift
@@ -22,6 +22,16 @@
 /// All input file paths are relative to the working directory where the tool is invoked.
 /// Output file paths may be absolute or relative to the working directory.
 public struct SafeDIToolManifest: Codable, Sendable {
+	private enum CodingKeys: String, CodingKey {
+		case dependencyTreeGeneration
+		case mockGeneration
+		case configurationFilePaths
+		case mockConfigurationOutputFilePath
+		case additionalMocksToGenerate
+		case additionalInputFiles
+		case activeCompilationConditions
+	}
+
 	/// A mapping from an input Swift file to an output file.
 	public struct InputOutputMap: Codable, Sendable {
 		/// The path to the input Swift file containing one or more root `@Instantiable` declarations.
@@ -65,6 +75,9 @@ public struct SafeDIToolManifest: Codable, Sendable {
 	/// Only populated by the `scan` subcommand; ignored by `generate`.
 	public var additionalInputFiles: [String]
 
+	/// Active conditional compilation flags used while scanning SafeDI declarations.
+	public var activeCompilationConditions: [String]
+
 	public init(
 		dependencyTreeGeneration: [InputOutputMap],
 		mockGeneration: [InputOutputMap] = [],
@@ -72,6 +85,7 @@ public struct SafeDIToolManifest: Codable, Sendable {
 		mockConfigurationOutputFilePath: String? = nil,
 		additionalMocksToGenerate: [String] = [],
 		additionalInputFiles: [String] = [],
+		activeCompilationConditions: [String] = [],
 	) {
 		self.dependencyTreeGeneration = dependencyTreeGeneration
 		self.mockGeneration = mockGeneration
@@ -79,5 +93,19 @@ public struct SafeDIToolManifest: Codable, Sendable {
 		self.mockConfigurationOutputFilePath = mockConfigurationOutputFilePath
 		self.additionalMocksToGenerate = additionalMocksToGenerate
 		self.additionalInputFiles = additionalInputFiles
+		self.activeCompilationConditions = activeCompilationConditions
+	}
+
+	public init(from decoder: Decoder) throws {
+		let container = try decoder.container(keyedBy: CodingKeys.self)
+		self.init(
+			dependencyTreeGeneration: try container.decode([InputOutputMap].self, forKey: .dependencyTreeGeneration),
+			mockGeneration: try container.decodeIfPresent([InputOutputMap].self, forKey: .mockGeneration) ?? [],
+			configurationFilePaths: try container.decodeIfPresent([String].self, forKey: .configurationFilePaths) ?? [],
+			mockConfigurationOutputFilePath: try container.decodeIfPresent(String.self, forKey: .mockConfigurationOutputFilePath),
+			additionalMocksToGenerate: try container.decodeIfPresent([String].self, forKey: .additionalMocksToGenerate) ?? [],
+			additionalInputFiles: try container.decodeIfPresent([String].self, forKey: .additionalInputFiles) ?? [],
+			activeCompilationConditions: try container.decodeIfPresent([String].self, forKey: .activeCompilationConditions) ?? [],
+		)
 	}
 }
--- a/Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
+++ b/Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
@@ -1,0 +1,347 @@
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import SwiftSyntax
+
+/// Selects the `#if` branch SafeDITool should scan without pretending to know
+/// compiler flags the caller did not provide.
+struct ConditionalCompilationEvaluator {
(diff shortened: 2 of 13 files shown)


Sources/SafeDICore/Models/SafeDIToolManifest.swift
```diff
@@ -22,6 +22,16 @@
...
+	private enum CodingKeys: String, CodingKey {
+		case dependencyTreeGeneration
+		case mockGeneration
+		case configurationFilePaths
+		case mockConfigurationOutputFilePath
+		case additionalMocksToGenerate
+		case additionalInputFiles
+		case activeCompilationConditions
+	}
+
```
dfed, line 34: we can remove. build artifacts are hermetic per SafeDI version. no need for custom decoding. breaking changes to the format are fine.
```suggestion
```

Sources/SafeDICore/Models/SafeDIToolManifest.swift
```diff
@@ -65,19 +75,37 @@ public struct SafeDIToolManifest: Codable, Sendable {
...
+		let container = try decoder.container(keyedBy: CodingKeys.self)
+		self.init(
+			dependencyTreeGeneration: try container.decode([InputOutputMap].self, forKey: .dependencyTreeGeneration),
+			mockGeneration: try container.decodeIfPresent([InputOutputMap].self, forKey: .mockGeneration) ?? [],
+			configurationFilePaths: try container.decodeIfPresent([String].self, forKey: .configurationFilePaths) ?? [],
+			mockConfigurationOutputFilePath: try container.decodeIfPresent(String.self, forKey: .mockConfigurationOutputFilePath),
+			additionalMocksToGenerate: try container.decodeIfPresent([String].self, forKey: .additionalMocksToGenerate) ?? [],
+			additionalInputFiles: try container.decodeIfPresent([String].self, forKey: .additionalInputFiles) ?? [],
+			activeCompilationConditions: try container.decodeIfPresent([String].self, forKey: .activeCompilationConditions) ?? [],
+		)
```
dfed, line 109: same as above
```suggestion
```

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+
+			switch evaluate(condition) {
+			case let .known(isActive):
+				if isActive {
+					return .active(clause)
+				}
+			case let .unknown(error):
+				return .unevaluable(error)
+			}
+		}
```
dfed, line 42: can we use a `first(where:` clause to avoid early returns from `if` and `switch` statements? we only early return from `guard` in this repo

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+
+		if let tuple = TupleExprSyntax(condition),
+		   tuple.elements.count == 1,
+		   let element = tuple.elements.first
+		{
+			return evaluate(element.expression)
+		}
+
+		return .unknown(.init(condition: condition.trimmedDescription))
+	}
```
dfed, line 97: free-handed this so may not be right (hopefully tests will catch if I screwed up!), but let's restructure to avoid early returns:
```suggestion
		if let boolLiteral = BooleanLiteralExprSyntax(condition) {
			return .known(boolLiteral.literal.tokenKind == .keyword(.true))
		} else if let identifier = DeclReferenceExprSyntax(condition) {
			let name = identifier.baseName.text
			if activeCompilationConditions.contains(name) {
				return .known(true)
			} else {
				// Missing custom flags are unknown, not false. The scanner can be
				// invoked outside the compiler command that ultimately defines `-D`
				// flags, so silently treating an absent flag as inactive would hide
				// guarded SafeDI declarations from generation.
				return .unknown(.init(condition: name, suggestedActiveCondition: name))
			}
		} else if let prefixOperator = PrefixOperatorExprSyntax(condition) {
			if prefixOperator.operator.text == "!" {
				return evaluate(prefixOperator.expression).negated	
			} else {
				return .unknown(.init(condition: condition.trimmedDescription))
			}
		} else if let sequence = SequenceExprSyntax(condition) {
			return evaluateSequence(Array(sequence.elements), fallbackDescription: condition.trimmedDescription)
		} else if let infixOperator = InfixOperatorExprSyntax(condition),
		   let binaryOperator = BinaryOperatorExprSyntax(infixOperator.operator)
		{
			let left = evaluate(infixOperator.leftOperand)
			let right = evaluate(infixOperator.rightOperand)
			switch binaryOperator.operator.text {
			case "&&":
				return left.and(right)
			case "||":
				return left.or(right)
			default:
				return .unknown(.init(condition: condition.trimmedDescription))
			}
		} else if let tuple = TupleExprSyntax(condition),
		   tuple.elements.count == 1,
		   let element = tuple.elements.first
		{
			return evaluate(element.expression)
		} else {
			return .unknown(.init(condition: condition.trimmedDescription))
		}
	}
```

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+			return false
+		}
+		index += 1
+		return true
+	}
+}
+
+enum ConditionalCompilationClauseSelection {
+	case active(IfConfigClauseSyntax?)
+	case unevaluable(UnevaluableConditionalCompilationConditionError)
```
dfed, line 159: lazy q: is this actually reachable? and do we have tests that cover it?

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+		switch (self, other) {
+		case (.known(false), _), (_, .known(false)):
+			.known(false)
+		case (.known(true), let other):
+			other
+		case (let selfEvaluation, .known(true)):
+			selfEvaluation
+		case (.unknown, _):
+			self
+		}
```
dfed, line 185: I clearly need to update our linter! would love to lift the `let`s:

```suggestion
		switch (self, other) {
		case (.known(false), _), (_, .known(false)):
			.known(false)
		case let (.known(true), other):
			other
		case let (selfEvaluation, .known(true)):
			selfEvaluation
		case (.unknown, _):
			self
		}
```

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+		case (.known(true), _), (_, .known(true)):
+			.known(true)
+		case (.known(false), let other):
+			other
+		case (let selfEvaluation, .known(false)):
+			selfEvaluation
+		case (.unknown, _):
+			self
+		}
+	}
```
dfed, line 199: same here with the `let` extraction

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+		switch (self, other) {
+		case (.known(true), _), (_, .known(true)):
+			.known(true)
+		case (.known(false), let other):
+			other
+		case (let selfEvaluation, .known(false)):
+			selfEvaluation
+		case (.unknown, _):
+			self
+		}
```
dfed, line 198: this is inspired. I really like how you structured this!

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+	override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+		visitDecl(node)
+	}
+
+	override func visit(_ node: ExtensionDeclSyntax) -> SyntaxVisitorContinueKind {
+		if node.attributes.instantiableMacro != nil {
+			containsSafeDIDeclaration = true
+			return .skipChildren
+		}
+		return containsSafeDIDeclaration ? .skipChildren : .visitChildren
```
dfed, line 257: here and below let's add `else` branches:

```suggestion
		} else {
			return containsSafeDIDeclaration ? .skipChildren : .visitChildren
		}
```

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+
+private final class InstantiableBodySyntaxDetector: SyntaxVisitor {
+	init(customMockName: String?) {
+		self.customMockName = customMockName
+		super.init(viewMode: .sourceAccurate)
+	}
+
+	var containsInstantiableBodySyntax = false
+
+	private let customMockName: String?
```
dfed, line 298: `private` below `internal` please per repo convention

Sources/SafeDICore/Visitors/ConditionalCompilationEvaluator.swift
```diff
@@ -0,0 +1,347 @@
...
+			walk(switchCases)
+		case let .decls(decls):
+			walk(decls)
+		case let .postfixExpression(expression):
+			walk(expression)
+		case let .attributes(attributes):
+			walk(attributes)
+		}
+	}
+}
```
dfed, line 217: help me understand why we need this? can't we just `visitChildren` rather than explicitly walking the elements within the `ifConfig` manually?

dfed (commented):
Very nifty! I've wanted this for awhile, but haven't carved out the time to get it over the line. Very much a fan of this capability. There are a few things we'll need before this gets approved (beyond handling the inline feedback):

* patch test coverage **must** be at 100%. the test suite should prove that this works.
* The SafeDITool plugins (SPM + Xcode, as well as the `safedi.bzl` and Tuist `SafeDI.swift` plugins) should be updated to gather this information and pass it to the tool commands. This capability is not terribly helpful unless the vended plugins actually pass this data into the into the tool.
* ideally we'd take a `mockOnly: true` in our example projects and put it behind `#if DEBUG` to prove the system works end-to-end.

Happy to answer any questions you have. Also happy to chat over in iOS Folks Slack if that's easier.

adam-zethraeus:
Thanks for the feedback!
I'm on UK time at the moment but I expect to be able to take a pass on this tomorrow — and see if I can catch you on slack.

